PR #4, opened 2018-10-12
Appends / Fixes #1

### LINKED ISSUE #1: Feature - Inventory System Implementation

EtiTheSpirit:
Appended the methods to teleport without changing worlds / recreating player object, these are not active in Player due to the lack of client replication for now.

Made the ReadSingle/ReadDouble methods of PacketReader and the Write Float/Double methods of PacketWriter safe via the usage of System.BitConverter


Source/NexusForever.WorldServer/Game/Map/MapManager.cs
```diff
@@ -10,47 +10,66 @@
...
-                tasks.Add(Task.Run(() => { map.Update(lastTick); }));
-
-            Task.WaitAll(tasks.ToArray());
-
-            sw.Stop();
-            if (sw.ElapsedMilliseconds > 10)
-                log.Warn($"{maps.Count} map(s) took {sw.ElapsedMilliseconds} to update!");
-        }
-    }
+	public static class MapManager
```
Drutol: Should such components be statics? I see quite a few of them and I wonder if they shouldn't be injected into stuff as dependencies for example.

No clue about architecture or whether it's temporary, just my 2 cents ^^

Source/NexusForever.WorldServer/Network/Message/Handler/SocialHandler.cs
```diff
@@ -1,19 +1,31 @@
...
+using NLog;
+using System;
 
 namespace NexusForever.WorldServer.Network.Message.Handler
 {
     public static class SocialHandler
     {
+        private static readonly Logger log = LogManager.GetCurrentClassLogger();
+
+        public static readonly string CommandPrefix = "!";
```
EtiTheSpirit, line 16: I don't know if this is 100% necessary, I did it out of habit. In the case that commands are used anywhere else this would be more useful.

PR #17, opened 2018-11-29
Added Client Connector

MaxtorCoder:
And some other small stuffs, just ignore the ClientWhoRequest structure. 🔒


Source/NexusForever.ClientConnector/Program.cs
```diff
@@ -0,0 +1,36 @@
...
+            uint dwCreationFlags,
+            IntPtr lpEnvironment,
+            string lpCurrentDirectory,
+            ref STARTUPINFO lpStartupInfo,
+            out PROCESS_INFORMATION lpProcessInformation
+        );
+
+        static void Main(string[] args)
+        {
+            Console.WriteLine("Type in your host name : ");
```
lucas-jones: Maybe the arguments could be loaded from a text / `.json`. Which if it doesn't exist will be created with the default of `localhost`. 

This way users won't have to keep inputting the host name every time they want to play :)

Source/NexusForever.WorldServer/Network/Message/Model/ClientWhoRequest.cs
```diff
@@ -6,8 +6,41 @@ namespace NexusForever.WorldServer.Network.Message.Model
     [Message(GameMessageOpcode.ClientWhoRequest, MessageDirection.Client)]
     public class ClientWhoRequest : IReadable
     {
+        #region off_DB3184
```
lucas-jones: Accidental :P

lucas-jones (commented):
I've added a few comments

Source/NexusForever.ClientConnector/Program.cs
MaxtorCoder: Good plan, thanks :D

Source/NexusForever.WorldServer/Network/Message/Model/ClientWhoRequest.cs
MaxtorCoder: What do you mean? :P

PR #18, opened 2018-11-29
OS independent directory separator, all game tables added as comment

ahoiahoi:
tested on linux & windows


Fabi:
The `/` seperator is fully window compatible.
If you use that one you can use $`"tbl{Path.DirectorySeparatorChar}CharacterCreation.tbl"` as alternative syntax

ahoiahoi:
thanks Fabi - did not know that

Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
```diff
@@ -45,6 +46,8 @@ public static async Task<Account> GetAccountAsync(string email, byte[] sessionKe
         /// </summary>
         public static void CreateAccount(string email, string password)
         {
+            if (!new EmailAddressAttribute().IsValid(email))
```
Fabi: actually that's not the case. The SRP6a implementation needs to be re checked ^^

PR #19, opened 2018-11-30
Wrap commands in blanket try catch

Matt-Hurd:
Exceptions from server commands definitely should not crash the server. While each command should have it's own error handling, wrapping them all is a (barely) acceptable precaution to ease development.

Also removing the existing blanket try/catch blocks, as they are no longer needed and add extra bulk.

### DIFF
--- a/Source/NexusForever.WorldServer/WorldServer.cs
+++ b/Source/NexusForever.WorldServer/WorldServer.cs
@@ -67,7 +67,14 @@ namespace NexusForever.WorldServer
 
                 CommandHandlerDelegate handler = CommandManager.GetCommandHandler(command);
                 if (handler != null)
-                    handler.Invoke(null, parameters);
+                    try
+                    {
+                        handler.Invoke(null, parameters);
+                    }
+                    catch (Exception exception)
+                    {
+                        log.Error(exception);
+                    }
                 else
                     Console.WriteLine("Invalid command!");
             }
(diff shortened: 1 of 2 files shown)


Source/NexusForever.WorldServer/WorldServer.cs
```diff
@@ -67,7 +67,14 @@ private static void Main()
 
                 CommandHandlerDelegate handler = CommandManager.GetCommandHandler(command);
                 if (handler != null)
-                    handler.Invoke(null, parameters);
```
Rawaho, line 70: I would prefer blocks with more than a single line of code to be wrapped in curly braces, the else below is fine since it's just a single line.

Rawaho (changes requested):

Source/NexusForever.WorldServer/WorldServer.cs
Matt-Hurd: Am I blind? I only see one line here.

Rawaho: The code that has replaced this removed line (the try catch) needs to wrapped in curly braces.

Matt-Hurd: Was looking on the wrong branch on local 😢

Rawaho (approved):

PR #32, opened 2018-12-01
Connection string changes

jasoncouture:
In preparation for other work I have done, I have adjusted how the connection string configuration works to allow multiple providers to be used. This also enables using either the previous configuration for MySql, or using a straight connection string instead.


jasoncouture:
Replaced by #60

PR #33, opened 2018-12-01
[WIP] Add vendor and currency support

Matt-Hurd:
Should've made more commits than this, ugh.

Implemented:
* Add full implementation of currencies
* Basic Vendor buy/sell, manipulating items and currencies

Unimplemented:
* Most items don't have explicit values set and are instead calculated based on quality, item type, etc. This equation will need to be figured out.
* Buyback from vendors isn't set
* Reason for add/delete from inventory is not being set to the proper value

### DIFF
--- a/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
+++ b/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
@@ -12,7 +12,7 @@
         ServerChangeWorld               = 0x00AD,
         ServerBuybackItemsUpdated       = 0x00BA,
         ClientBuybackItemFromVendor     = 0x00BB,
-        ServerBuybackItemRemoved         = 0x00BD,
+        ServerBuybackItemRemoved        = 0x00BD,
         ClientVendorPurchase            = 0x00BE,
         ClientCharacterLogout           = 0x00BF,
         ServerCharacterCreate           = 0x00DC,
--- a/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
+++ b/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
@@ -13,6 +13,7 @@
         ClientVendorPurchase            = 0x00BE,
         ClientCharacterLogout           = 0x00BF,
         ServerCharacterCreate           = 0x00DC,
+        ServerCurrencyGainText          = 0x00DD,
         Server00F1                      = 0x00F1,
         Server0104                      = 0x0104, // Galactic Archive
         ServerCharacter                 = 0x010F, // single character
@@ -22,6 +23,7 @@
         ServerItemDelete                = 0x0148,
         ClientItemDelete                = 0x0149,
         ServerCharacterSelectFail       = 0x0162,
+        ClientVendorSell                = 0x0166,
         Server0169                      = 0x0169, // ability book related
         ClientItemSplit                 = 0x017D,
         ServerItemStackCountUpdate      = 0x017F,
@@ -72,6 +74,7 @@
         ServerEntityDestory             = 0x088C, // 0x8A5??
         Server08B3                      = 0x08B3,
         ServerVendor                    = 0x090B,
+        ServerCurrencySet               = 0x0919,
         ServerItemVisualUpdate          = 0x0933,
         Server0934                      = 0x0934,
         ServerEmote                     = 0x093C,
--- a/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
+++ b/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
@@ -23,7 +23,7 @@
         ServerItemDelete                = 0x0148,
         ClientItemDelete                = 0x0149,
         ServerCharacterSelectFail       = 0x0162,
-        ClientVendorSell                = 0x0166,
+        ClientSellItemToVendor          = 0x0166,
         Server0169                      = 0x0169, // ability book related
         ClientItemSplit                 = 0x017D,
         ServerItemStackCountUpdate      = 0x017F,
--- a/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
+++ b/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
@@ -74,7 +74,7 @@
         ServerEntityDestory             = 0x088C, // 0x8A5??
         Server08B3                      = 0x08B3,
         ServerVendor                    = 0x090B,
-        ServerCurrencySet               = 0x0919,
+        ServerPlayerCurrencyChanged     = 0x0919,
         ServerItemVisualUpdate          = 0x0933,
         Server0934                      = 0x0934,
         ServerEmote                     = 0x093C,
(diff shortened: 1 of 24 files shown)


Source/NexusForever.WorldServer/Network/Message/Handler/VendorHandler.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            uint itemId = 0;
+            for (int x = 0; x < VendorInfo.Items.Count; x++)
+            {
+                if (VendorInfo.Items[x].Index == vendorPurchase.VendorIndex)
+                {
+                    itemId = VendorInfo.Items[x].ItemId;
+                    break;
+                }
+            }
```
Rawaho: Would be better if this was encapsulated into say VendorInfo

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -30,6 +30,7 @@ public class Player : WorldEntity, ISaveCharacter
         public Race Race { get; }
         public Class Class { get; }
         public List<float> Bones { get; }
+        public ulong[] Currencies { get;  }
```
Rawaho: Would be better if this was encapsulated having a bunch of functions that can modify currencies along with checking for valid currencies, amounts, overflows, ect...

Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
```diff
@@ -13,6 +13,7 @@ public enum GameMessageOpcode
         ClientVendorPurchase            = 0x00BE,
         ClientCharacterLogout           = 0x00BF,
         ServerCharacterCreate           = 0x00DC,
+        ServerCurrencyGainText          = 0x00DD,
```
Rawaho: Server packet names should based off any LUA events the handler in the client may call while processing.
In the case of 0x00DD it happens to call ChannelUpdate_Loot, maybe something like ServerChannelUpdateLoot would be better.

Source/NexusForever.WorldServer/Network/Message/Handler/VendorHandler.cs
```diff
@@ -0,0 +1,161 @@
...
+            if (itemId == 0)
+                return;
+
+            Item2Entry itemEntry = GameTableManager.Item.GetEntry(itemId);
+            float costMultiplier = vendorPurchase.VendorItemQty * VendorInfo.BuyPriceMultiplier;
+
+            if (session.Player.Currencies[itemEntry.CurrencyTypeId0] < (itemEntry.CurrencyAmount0 * costMultiplier))
```
Rawaho: Adding to the comment from before, it would be better if this was encapsulated with something like GetCurrency(123).

Source/NexusForever.WorldServer/Network/Message/Handler/VendorHandler.cs
```diff
@@ -0,0 +1,161 @@
...
+                return;
+
+            if (session.Player.Currencies[itemEntry.CurrencyTypeId1] < (itemEntry.CurrencyAmount1 * costMultiplier))
+                return;
+
+            if (itemEntry.CurrencyTypeId0 > 0)
+            {
+                session.EnqueueMessageEncrypted(new ServerCurrencySet
+                {
+                    CurrencyId = (byte)((itemEntry.CurrencyTypeId0 & 0xFF) - 1),
```
Rawaho: You sure of ((itemEntry.CurrencyTypeId0 & 0xFF) - 1)?

Source/NexusForever.WorldServer/Network/Message/Model/ServerCurrencyGainText.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public byte CurrencyId { get; set; }
+        public long Count { get; set; }
+
+        public void Write(GamePacketWriter writer)
+        {
+            writer.Write((CurrencyId & 0b1000) != 0);
+            writer.Write((CurrencyId & 0b0100) != 0);
+            writer.Write((CurrencyId & 0b0010) != 0);
+            writer.Write((CurrencyId & 0b0001) != 0);
```
Rawaho: This can be writer.Write(CurrencyId , 4);
The only issues I have though is another packet writes currency as 5 bits and this one 4, you sure this is actually the same type?

Rawaho (changes requested):
An awesome start, just a few issues that need correcting.
I have enjoyed watching you stream this, the offer still stands from Discord :)

Source/NexusForever.WorldServer/Network/Message/Model/ServerCurrencyGainText.cs
Matt-Hurd: Nope! The ordering of the currencies is even different than it is in the DB.

Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
Matt-Hurd: Resolved for this opcode, will get names for other opcodes soon.

Source/NexusForever.WorldServer/Network/Message/Model/ServerCurrencySet.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class ServerCurrencySet : IWritable
+    {
+        public byte CurrencyId { get; set; }
+        public ulong Count { get; set; }
+        public uint Unknown1 { get; set; } = 0;
+        public uint Unknown2 { get; set; } = 0;
+
+        public void Write(GamePacketWriter writer)
+        {
+            writer.Write(CurrencyId - 1, 5);
```
Matt-Hurd, line 16: Still not sure about this...

Matt-Hurd: But it works ¯\\\_(ツ)\_/¯

Source/NexusForever.WorldServer/Network/Message/Handler/VendorHandler.cs
```diff
@@ -0,0 +1,126 @@
...
+                return;
+
+            session.Player.Inventory.ItemDelete(vendorSell.ItemLocation);
+
+            float sellMultiplier = VendorInfo.SellPriceMultiplier * vendorSell.Count;
+            Currency currency0 = session.Player.CurrencyManager.GetCurrency(itemEntry.CurrencyTypeId0);
+            Currency currency1 = session.Player.CurrencyManager.GetCurrency(itemEntry.CurrencyTypeId1);
+
+            if (currency0 != null)
+                session.Player.CurrencyManager.CurrencyAddCount((byte)itemEntry.CurrencyTypeId0, (uint)(itemEntry.CurrencyAmount0SellToVendor * sellMultiplier));
```
Matt-Hurd: Going to get these to ~100 characters long

Matt-Hurd:
This PR requires the GameTable fix, not included because it's out of scope.

Matt-Hurd:
Splitting into multiple PRs for easier maintainability

Issue #20, opened 2018-11-30
Implement "help" server command


n30na:
A command that provides the client with a list of possible commands.

jasoncouture:
@Rawaho I will do this either in #34 or a new PR if you take my checkpoint. I don't want to get too far out of sync with upstream (you) though.

PR #22, opened 2018-11-30
Some polishing on !itemadd and CommandHandlerAttribute

Dekker3D:
CommandHandlerAttribute can now be used multiple times on a method (for aliases like itemadd/additem). Itemadd now allows you to leave out the amount, assuming it to be one. It simply returns if there is no item ID.

PR #34, opened 2018-12-01
ASP.NET Core support

jasoncouture:
I have initialized ASP.NET Core. And moved commands over to a dependency injection style to allow them to be shared by the command line console, web, /c (cheat), and chat.
I will submit a seperate PR for the web stuff when I get to that.


jasoncouture:
d1d9808 fixes #20 and implements some of the polishing done in #22

PR #35, opened 2018-12-02
Heartbeat/clientping workaround

ahoiahoi:
workaround for now to prevent the 300s server disconnect


jasoncouture:
@ahoiahoi for the game table stuff, wouldn't it really be better to use a factory and Path.Combine() so the table location could be set in the config?

jasoncouture:
@ahoiahoi something like `GameTable<T> GameTableFactory.Load<T>(string fileName) { ... }`

ahoiahoi:
the gametable stuff is already pulled - this PR is about https://github.com/Rawaho/NexusForever/pull/35/commits/2472206506f0fa69f9e298613fd2773f2ad82d2c

PR #36, opened 2018-12-02
Misc improvements, including web socket console

jasoncouture:
Console accessible at http://localhost:5000/console.html

Added resharper definitions, and used them to cleanup all command files.
Use IConfiguration instead of JSON.NET Directly.
For security reasons, encryption should use RNGCryptoServiceProvider instead of Random (Fixed)
Moved configuration builder logic to SharedConfiguration so that it doesn't clutter ConfigurationManager<T>


jasoncouture:
@Rawaho please close this, I will submit a new one. I acidentially pushed some changes I didn't mean to.

PR #38, opened 2018-12-03
Currency implementation

Matt-Hurd:
Follow-up PR: https://github.com/Matt-Hurd/NexusForever/pull/1

~Requires #39~


Matt-Hurd:
Commits got fucky on merge, should've rebased instead :(

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -175,6 +178,15 @@ private void SendPacketsAfterAddToMap()
...
             Session.EnqueueMessageEncrypted(playerCreate);
+
+            foreach (Currency currency in CurrencyManager)
+            {
+                Session.EnqueueMessageEncrypted(new ServerPlayerCurrencyChanged
+                {
+                    CurrencyId = (byte)currency.Id,
+                    Amount = currency.Amount,
+                });
+            }
```
Rawaho: Wouldn't it be better to send this in a single packet ServerPlayerCreate?

Source/NexusForever.WorldServer/Network/Message/Model/ServerCurrencyChanged.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class ServerPlayerCurrencyChanged : IWritable
+    {
+        public byte CurrencyId { get; set; }
+        public ulong Amount { get; set; }
+        public uint Unknown1 { get; set; } = 0;
+        public uint Unknown2 { get; set; } = 0;
+
+        public void Write(GamePacketWriter writer)
+        {
+            writer.Write(CurrencyId - 1, 5);
```
Rawaho: I'm still not certain on this, it just doesn't look right.
If it works for the time being I see no reason to stop the PR because of it but it's something I will look into at some point.

Rawaho (changes requested):

Source/NexusForever.WorldServer/Game/Entity/Player.cs
Matt-Hurd: Oh, I forgot that was in there. Yep.

Matt-Hurd: It seems like sending it in ServerPlayerCreate doesn't update the client. Is it possible that ServerPlayerCreate is incorrect?

Source/NexusForever.WorldServer/Network/Message/Model/ServerCurrencyChanged.cs
Matt-Hurd: Left a comment about it, I really can't figure out why this is.

Rawaho (approved):

PR #40, opened 2018-12-03
Fixed crashes when parameters are omitted.

IceReaper:
When creating an account, with too less parameters, the world server should not crash.

### DIFF
--- a/Source/NexusForever.WorldServer/Command/Handler/AccountCommandHandler.cs
+++ b/Source/NexusForever.WorldServer/Command/Handler/AccountCommandHandler.cs
@@ -16,6 +16,12 @@ namespace NexusForever.WorldServer.Command.Handler
         [SubCommandHandler("create", "email password - Create a new account")]
         public async Task HandleAccountCreate(CommandContext context, string subCommand, string[] parameters)
         {
+            if (parameters.Length < 2)
+            {
+                await SendHelpAsync(context).ConfigureAwait(false);
+                return;
+            }
+
             AuthDatabase.CreateAccount(parameters[0], parameters[1]);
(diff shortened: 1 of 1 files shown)


Rawaho:
Might be better to have some global solution for this, maybe in the attribute have expected parameter amount which is checked before invoke?

Source/NexusForever.WorldServer/Command/Handler/AccountCommandHandler.cs
```diff
@@ -16,6 +16,9 @@ public AccountCommandHandler()
         [SubCommandHandler("create", "email password - Create a new account")]
         public async Task HandleAccountCreate(CommandContext context, string subCommand, string[] parameters)
         {
+            if (parameters.Length < 2)
+                return;
```
jasoncouture, line 22: Instead of just returning, you should call await SendHelpAsync(context) first, to show the help text to the user.

Source/NexusForever.WorldServer/Command/Handler/AccountCommandHandler.cs
```diff
@@ -24,6 +27,9 @@ await context.SendMessageAsync($"Account {parameters[0]} created successfully")
         [SubCommandHandler("delete", "email - Delete an account")]
         public async Task HandleAccountDeleteAsync(CommandContext context, string subCommand, string[] parameters)
         {
+            if (parameters.Length < 1)
+                return;
```
jasoncouture, line 36: See previous comment about await SendHelpAsync(context);

jasoncouture (changes requested):

jasoncouture (approved):
Looks good.

PR #41, opened 2018-12-04
Linux sts fix

ahoiahoi:
Enforce appendance of a carriage return on linux


Source/NexusForever.WorldServer/Game/Entity/Static/Stat.cs
```diff
@@ -5,5 +5,6 @@ public enum Stat
         Health  = 0,
         Dash  = 9,
         Level = 10,
+	Shield = 20,
```
hyena, line 8:
```suggestion
        Shield = 20,
```

Source/NexusForever.Shared/Network/GameSession.cs
```diff
@@ -146,6 +146,9 @@ private void HandlePacket(ClientGamePacket packet)
                 && packet.Opcode != GameMessageOpcode.ClientEntityCommand)
                 log.Trace($"Received packet {packet.Opcode}(0x{packet.Opcode:X}).");
 
+            //FIXME workaround for now...
```
hyena, line 149: Explaining what problem this workaround is solving will help another developer (or Future You) fix it properly.

ahoiahoi: unfortunately, this commit is from another branch & PR. 
I will re-create the PRs (just fyi, the server disconnects clients after 300s because no heatbeat is received by the clientping opcode)

PR #44, opened 2018-12-04
Linux STS Server fix

ahoiahoi:
Enforce appendance of a carriage return on linux - this is done by default on windows and the client expects the header to be cr/lf'ed


Fabi:
You can also force it for the whole stream by setting the `stream.NewLine` value

PR #48, opened 2018-12-05
Properly Save Deleted Items

Matt-Hurd:
This still doesn't seem to work, even though it is properly reaching the point to remove it from the context...


Matt-Hurd:
Gah, I figured `List` would have handlers for that. I need to stop assuming that C# is a magic language.

PR #51, opened 2018-12-06
Enum: Path - Added correct numerical values

kirmmin:


Fabi:
Result is the same tho :D

kirmmin:
Agreed, but for me, it's easier to cross-reference IDs with the database when they're in plain sight like this. I dunno, I may be alone, I just thought it would be clearer for other devs which ID = which path* :)

kirmmin:
Also, forgot to mention that some of the Enums begin with 1 e.g. https://github.com/Rawaho/NexusForever/blob/master/Source/NexusForever.WorldServer/Game/Entity/Static/Race.cs. Whereas others begin with 0 (like this one), just wanted to prevent confusion :)

PR #54, opened 2018-12-06
Support for different client languages.

IceReaper:
Added choosable language for the ClientConnector.
Additionally renamed Unknown22C to Language as it contains the id for the currently selected language.
The id matches the entries in the Language game table.

### DIFF
--- a/Source/NexusForever.AuthServer/Network/Message/Model/ClientHelloAuth.cs
+++ b/Source/NexusForever.AuthServer/Network/Message/Model/ClientHelloAuth.cs
@@ -1,4 +1,5 @@
-﻿using NexusForever.Shared.Network;
+using NexusForever.Shared.GameTable.Static;
+using NexusForever.Shared.Network;
 using NexusForever.Shared.Network.Message;
 
 namespace NexusForever.AuthServer.Network.Message.Model
@@ -61,7 +62,7 @@ namespace NexusForever.AuthServer.Network.Message.Model
         public NetworkGuid Unknown208 { get; } = new NetworkGuid();
         public NetworkGuid GameToken { get; } = new NetworkGuid();
         public uint Unknown228 { get; private set; }
-        public uint Unknown22C { get; private set; }
+        public Language Language { get; private set; }
         public uint Unknown230 { get; private set; }
         public uint Unknown234 { get; private set; }
         public HardwareInformation Hardware { get; } = new HardwareInformation();
@@ -77,7 +78,7 @@ namespace NexusForever.AuthServer.Network.Message.Model
             GameToken.Read(reader);
 
             Unknown228 = reader.ReadUInt();
-            Unknown22C = reader.ReadUInt();
+            Language = reader.ReadEnum<Language>(32u);
             Unknown230 = reader.ReadUInt();
             Unknown234 = reader.ReadUInt();
 
--- a/Source/NexusForever.ClientConnector/Program.cs
+++ b/Source/NexusForever.ClientConnector/Program.cs
@@ -31,30 +31,30 @@ namespace NexusForever.ClientConnector
                 Console.Write("Type in your host name: ");
                 string hostName = Console.ReadLine();
 
+                Console.Write("Type in your language: [en,de]");
+                string language = Console.ReadLine();
+
                 ClientConfiguration clientConfig = new ClientConfiguration
                 {
-                    HostName = hostName
+                    HostName = hostName,
+                    Language = language
                 };
 
                 File.WriteAllText(jsonFile, JsonConvert.SerializeObject(clientConfig));
+            }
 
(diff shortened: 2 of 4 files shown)


Source/NexusForever.Shared/GameTable/Static/Language.cs
```diff
@@ -3,7 +3,7 @@
     public enum Language
     {
         English = 1,
-        German,
+        German = 2,
```
jasoncouture: Do we need to be explicit with sequential values? = 2 does seem more obvious, but the other code that uses enums does not do this.

jasoncouture:
```suggestion
        German,
```

Source/NexusForever.Shared/GameTable/Static/Language.cs
```diff
@@ -3,7 +3,7 @@
     public enum Language
```
jasoncouture, line 3:
```suggestion
     public enum Language : uint
```

Source/NexusForever.AuthServer/Network/Message/Model/ClientHelloAuth.cs
```diff
@@ -61,7 +61,7 @@ public void Read(GamePacketReader reader)
         public NetworkGuid Unknown208 { get; } = new NetworkGuid();
         public NetworkGuid GameToken { get; } = new NetworkGuid();
         public uint Unknown228 { get; private set; }
-        public uint Unknown22C { get; private set; }
+        public uint Language { get; private set; }
```
jasoncouture:
```suggestion
        public Language Language { get; private set; }
```

Source/NexusForever.AuthServer/Network/Message/Model/ClientHelloAuth.cs
```diff
@@ -77,7 +77,7 @@ public void Read(GamePacketReader reader)
             GameToken.Read(reader);
 
             Unknown228 = reader.ReadUInt();
-            Unknown22C = reader.ReadUInt();
+            Language = reader.ReadUInt();
```
jasoncouture:
```suggestion
            Language = reader.ReadEnum<Language>(16u);
```

Source/NexusForever.ClientConnector/Program.cs
```diff
@@ -31,30 +31,34 @@ public static void Main()
...
                 File.WriteAllText(jsonFile, JsonConvert.SerializeObject(clientConfig));
 
-                LaunchClient(hostName);
+                LaunchClient(hostName, language);
             }
             else
             {
                 ConfigurationManager<ClientConfiguration>.Initialise(jsonFile);
-                LaunchClient(ConfigurationManager<ClientConfiguration>.Config.HostName);
+                LaunchClient(ConfigurationManager<ClientConfiguration>.Config.HostName, ConfigurationManager<ClientConfiguration>.Config.Language);
```
jasoncouture: Why not just pass the configuration down to LaunchClient, instead of adding parameters. There will be more in the future. I'd like to be able to change the realm datacenter ID to see what logic that triggers in the client. May be worth considering.

Source/NexusForever.Shared/GameTable/Static/Language.cs
IceReaper: As my Language.tbl list 1,2,3,5 we might need  this for the future. I did decide to not add 3 for French, as i want to explicitly test this before adding it and im still lacking the french locale. 5 seems to be Chinese (Language.tbl), not sure where this Korean enum value comes from?

Source/NexusForever.AuthServer/Network/Message/Model/ClientHelloAuth.cs
IceReaper: Shouldn't this be 32u then?

Source/NexusForever.Shared/GameTable/Static/Language.cs
jasoncouture: Ported from WildStar studio. If more than 1 is numbered, maybe instead of the previous suggestion, they should all be numbered?

I don't think the KR or CN releases went out. I believe only 1/2/3 are valid.

Source/NexusForever.AuthServer/Network/Message/Model/ClientHelloAuth.cs
jasoncouture: Yup. My brain for some reason read that as UShort lol.

Source/NexusForever.Shared/GameTable/Static/Language.cs
jasoncouture: In language.tbl, do you have localised text id? We can look those up if you do.

PR #55, opened 2018-12-06
Debug packet.

IceReaper:
Usefull for ingame testing of packet data.
Values can be defined as decimal numbers, binary numbers using a ```0b``` prefix or hexadecimal numbers using ```0x``` prefix. Additionaly the suffix ```|123``` can be used to specify the amount of send bits. Also i added a value ```PLAYER``` which references the current players guid.

Example to make the character dance:
```
!debug packet 093C PLAYER|32 7|4 34|14
```

### DIFF
--- a/Source/NexusForever.Shared/Network/GameSession.cs
+++ b/Source/NexusForever.Shared/Network/GameSession.cs
@@ -17,7 +17,7 @@ namespace NexusForever.Shared.Network
         /// </summary>
         public bool CanProcessPackets { get; set; } = true;
 
-        protected PacketCrypt encryption;
+        public PacketCrypt Encryption;
 
         private FragmentedBuffer onDeck;
         private readonly ConcurrentQueue<ClientGamePacket> incomingPackets = new ConcurrentQueue<ClientGamePacket>();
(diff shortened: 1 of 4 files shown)


Source/NexusForever.Shared/Network/GameSession.cs
```diff
@@ -17,7 +17,7 @@ public abstract class GameSession : NetworkSession
         /// </summary>
         public bool CanProcessPackets { get; set; } = true;
 
-        protected PacketCrypt encryption;
+        public PacketCrypt Encryption;
```
jasoncouture, line 20: This should probably not be public.

jasoncouture:
Awesome command, by the way!

Source/NexusForever.WorldServer/Command/Handler/DebugHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+                        else if (Regex.IsMatch(parts[0], "^0b[0-1]+$"))
+                            value = Convert.ToUInt64(parts[0].Substring(2), 2);
+                        else if (Regex.IsMatch(parts[0], "^0x[0-9A-F]+$"))
+                            value = Convert.ToUInt64(parts[0].Substring(2), 16);
+                        else if (parts[0] == "PLAYER")
+                            value = context.Session.Player.Guid;
+                        else
+                            continue;
+
+                        writer.Write(value, parts.Length > 1 ? Convert.ToUInt32(parts[1], 10) : 64);
```
hyena, line 48: I think it would be useful if the default weren't hardcoded here to 64 bits. Maybe it could simply send exactly as many bytes as the user typed when there's no `|`?

hyena: For example, I want to test packets that have contents like `18D74E00F4670E0000` or `18D74E000122000000002AADD745` and I don't yet understand their layout.

Rawaho:
This exposes too much publicly, my recommendation would be to have a function on the session that you pass a byte array or writer into which actually builds and enqueues internally.

Rawaho:
Can be reopened when the changes to the encryption publicity are made

PR #58, opened 2018-12-07
Initial support for logout from both client and server.

jasoncouture:
TODO: Event system will need to be reworked to enforce logout delay.


Source/NexusForever.WorldServer/Network/Message/Handler/ClientLogoutHandler.cs
```diff
@@ -0,0 +1,46 @@
+using NexusForever.Shared.Game.Events;
+using NexusForever.Shared.Network.Message;
+using NexusForever.WorldServer.Network.Message.Model;
+using NLog;
+
+namespace NexusForever.WorldServer.Network.Message.Handler
+{
+
+    public static class ClientLogoutHandler
```
MaxtorCoder: You should make this in the CharacterHandler. 😄

Rawaho: I do agree with this, I doubt we need an entire handler class for a single packet

Rawaho:
I think it would be best until the event system is reworked to remove the event entirely since it will probably need to be changed when the system is updated anyway.

Source/NexusForever.WorldServer/Network/Message/Handler/ClientLogoutHandler.cs
jasoncouture: I made both requested changes.

MaxtorCoder: Thanks you!

PR #62, opened 2018-12-07
Player packet identification

kirmmin:


Source/NexusForever.WorldServer/Game/Entity/Network/Model/PlayerEntityModel.cs
```diff
@@ -18,7 +18,7 @@ public class PlayerEntityModel : IEntityModel
         public byte Unknown34 { get; set; }
         public List<ulong> Unknown3C { get; } = new List<ulong>();
         public List<float> Bones { get; set; } = new List<float>();
-        public byte Unknown48 { get; set; }
+        public byte PvPEnabled { get; set; } // 0 = off, 1+ = on
```
Rawaho: Having an enum and calling this `PvpFlags` would be better as it has more than 2 values.
From the quick look I had in the client I had, 0x01 is called flagged, and 0x02 is called forced

Rawaho (changes requested):

Source/NexusForever.WorldServer/Game/Entity/Network/Model/PlayerEntityModel.cs
kirmmin: Gotcha :) I'll get working on that. I have some sniffs from another member, and one of them had 6 as the value in this field. Any chance you could tell me what 6 might mean?

kirmmin:
```
{ id: 453416705,
  realmId: 388,
  name: '<Name Redacted>',
  race: 'Mordesh',
  clazz: 'Stalker',
  sex: 'Male',
  Unknown20: 0,
  Unknown2C_Length: 0,
  Unknown2C: '',
  guildName: '<Guild Name Redacted>',
  unknown34: 5,
  Unknown3C_Length: 2,
  Unknown3C: '0xFF91 0xFEBA ',
  Bones_Length: 14,
  Bones: '0x0001 0x0000 0x0000 0x0000 0x0001 0x0000 0x0000 0x0000 0x0000 0x0000 0x0000 0x0000 0x0000 0x0000 ',
  PvP: 0,
  Unknown4C: 0,
  TitleId: 0 }
```
This is part of the sniff that was sent to me by @lucas-jones on Discord, who confirmed the `realmId` placement, possibly used for cross-server battles amongst other things.

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -119,12 +119,13 @@ protected override IEntityModel BuildEntityModel()
             return new PlayerEntityModel
             {
                 Id       = CharacterId,
-                Unknown8 = 358,
+                RealmId  = 358,
```
Matt-Hurd, line 136: 358 seems like a magic number here, can we extract it as a const?

kirmmin: As I understand it, it was the EU server Realm ID. I think we'd want to update this to be the realm that the user is actually connecting on.
I can dig through the code and see if it's exposed. I should think we can get a reference to it, or create one, either way.

Rawaho: This functionality already exists on the housing branch (housing does some weird stuff if the realm id doesn't match), no need to add an implementation for it

Source/NexusForever.WorldServer/Game/Entity/Network/Model/PlayerEntityModel.cs
kirmmin: This should be ready for your resolution & sign off, @Rawaho. Thanks.

Source/NexusForever.WorldServer/Game/Entity/Player.cs
kirmmin: Good to know. Thanks @Rawaho. Is this part of the PR fine to leave, then, as it'll get replaced by the housing branch? Alternatively, do you need me to remove the `RealmId` reference?

lucas-jones: 356 is US realm I believe. We should be using a consistent realmId otherwise it causes inconsistencies and the client been confused.

Even if the value is currently wrong, the naming is useful.

If Rawaho has a solution for this in the housing branch we can easily sort this :)

Source/NexusForever.WorldServer/Game/Entity/Static/PvPFlag.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace NexusForever.WorldServer.Game.Entity.Static
+{
+    public enum PvPFlag
+    {
+        Disabled    = 0,
+        Enabled     = 1,
+        Forced      = 2, // Seems to be forced to server settings. Toggle "on" is disabled on PvE server. Possibly a city thing?
+        Unknown     = 6 // Same deal as PvPFlag.Forced.
```
Rawaho: This doesn't need to be a member, it's just a combination of 2 flags, 2 | 4, also needs the Flags attribute

Source/NexusForever.WorldServer/Game/Entity/Static/Stat.cs
```diff
@@ -2,8 +2,33 @@
...
+        Resource4           = 6,    // Must be float. How do we enforce?
+        Resource5           = 7,    // (Assumed Stat) Must be float. How do we enforce?
+        Resource6           = 8,    // (Assumed Stat) Must be float. How do we enforce?
+        Dash                = 9,    // Must be float. How do we enforce?
+        Level               = 10,
+        MentorLevel         = 11,
+        StandState          = 12, // 0 = Standing (Combat Pose), 1 = Sitting, 2 = Laying Down, 3 = Standing (Idle) More info: https://github.com/Hammster/wildstar-api-docs/blob/36be999b77a9dcdc4b27e95d217d54e419fdcbf3/Classes/Unit.md#getstandstate
+        Sheathed            = 15, // 0 = Unsheathed Weapons, 1 = Sheathed Weapons
+        Shield              = 20,
+        // Stat(32) Duplicates Health, Stat(33) Duplicates Focus, and so on. Seems to loop at this point, suggesting a maximum of 32 "Stats".
```
Rawaho, line 32: Not sure what you mean here by "Must be float. How do we enforce?"

Rawaho (changes requested):

Source/NexusForever.WorldServer/Game/Entity/Static/Stat.cs
kirmmin: Some stats fail to work when they are `int` and will only work when sent in as `float`. Not sure how we enforce a 'stat type' e,.g. `int` or `float`, based on the Enum?

Source/NexusForever.WorldServer/Game/Entity/Static/PvPFlag.cs
kirmmin: Thank you. I will get this updated (and I'll try out `4` as well :p

kirmmin: `4` doesn't seem to do anything different. Are we assuming that there is a 4 but we don't know what it is yet, given that we have logs that show a `6`?

kirmmin: I removed it for now - we can add it back in, should that be necessary. The 4 states that I can see for PvP are available with the first 3 flags.

PR #63, opened 2018-12-07
Player faction updates

kirmmin:
Sets faction at creation
Applies factionId to character table

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Entity/Player.cs
+++ b/Source/NexusForever.WorldServer/Game/Entity/Player.cs
@@ -64,6 +64,7 @@ namespace NexusForever.WorldServer.Game.Entity
             Level       = model.Level;
             Bones       = new List<float>();
             CurrencyManager = new CurrencyManager(this, model);
+            Faction2    = model.FactionId;
 
             Inventory   = new Inventory(this, model);
             Session     = session;
@@ -161,7 +162,7 @@ namespace NexusForever.WorldServer.Game.Entity
             {
                 FactionData = new ServerPlayerCreate.Faction
                 {
-                    FactionId = 166,
+                    FactionId = 166, // This does not do anything for the player's "main" faction. Exiles/Dominion
                 }
             };
 
--- a/Source/NexusForever.WorldServer/Network/Message/Model/ServerPlayerCreate.cs
+++ b/Source/NexusForever.WorldServer/Network/Message/Model/ServerPlayerCreate.cs
@@ -24,7 +24,7 @@ namespace NexusForever.WorldServer.Network.Message.Model
             }
 
             public ushort FactionId { get; set; }
-            public List<FactionReputation> FactionReputations { get; set; } = new List<FactionReputation>();
+            public List<FactionReputation> FactionReputations { get; } = new List<FactionReputation>();
 
             public void Write(GamePacketWriter writer)
             {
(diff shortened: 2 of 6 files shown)


kirmmin:
I have questions around `ServerPlayerCreate.Faction` & `Faction1` in Entitiy Model.

`ServerPlayerCreate.Faction`: This seems to be tied to some other value, as it did not affect my character's "display" of faction or reputation values.
`EntityModel.Faction1` - Not sure what this is used for as Faction2 is clearly in relation to the player's actual faction value. Could be an NPC or 'faction usable' flag? It was also mentioned by neone, on Discord, that it could be in relation to regional factions.

Source/NexusForever.WorldServer/Network/Message/Model/ServerPlayerCreate.cs
```diff
@@ -24,7 +24,7 @@ public void Write(GamePacketWriter writer)
             }
 
             public ushort FactionId { get; set; }
-            public List<FactionReputation> FactionReputations { get; } = new List<FactionReputation>();
+            public List<FactionReputation> FactionReputations { get; set; } = new List<FactionReputation>();
```
jasoncouture: This should probably be get only, use .Add .Clear, and .AddRange to modify it.

kirmmin: Yea, this was me testing. I hard-set it and it wasn't working in the client, so I stopped updates of it until I've had chance to revisit it and understand that piece. I can remove that line of commit, too.

kirmmin:
This should be finalised and ready for merge. This PR focuses solely on setting up the character to be on the right faction, and extend that logic into the right classes.
I have started work on a reputation manager which will take on the rest of the faction piece for `Players`: https://github.com/kirmmin/NexusForever/tree/reputation-management

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -64,6 +64,7 @@ public Player(WorldSession session, Character model)
             Level       = model.Level;
             Bones       = new List<float>();
             CurrencyManager = new CurrencyManager(this, model);
+            Faction2    = model.FactionId;
```
Rawaho, line 67: What about existing characters? Will this crash the client since it will be set to 0?

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -161,7 +162,7 @@ private void SendPacketsAfterAddToMap()
             {
                 FactionData = new ServerPlayerCreate.Faction
                 {
-                    FactionId = 166,
+                    FactionId = 166, // This does not do anything for the player's "main" faction. Exiles/Dominion
```
Rawaho, line 165: Afaik this sets your faction in the character window, should be the same as the one in the database

Rawaho (changes requested):

Source/NexusForever.WorldServer/Game/Entity/Player.cs
kirmmin: It didn't, for me. It seemed to re-default to Dominion.

Source/NexusForever.WorldServer/Game/Entity/Player.cs
kirmmin: I thought the same thing, but didn't seem to do anything, which led me to figuring out `Faction2`'s use. I'm wondering if it's been misidentified, or was for something else. Trying other values in it, like `167`, `0` or `1` did nothing in my testing.

Rawaho (approved):

PR #65, opened 2018-12-09
Persistent Character location

ahoiahoi:
Store player location in the DB

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Entity/Player.cs
+++ b/Source/NexusForever.WorldServer/Game/Entity/Player.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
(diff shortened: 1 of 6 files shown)


Rawaho:
Location and world is already stored on every grid entity, there is no need to have them duplicated on the player too.
My suggestion would be using `OnRelocate` override to set the save mask for location and when the actual save occurs pull the location from the grid entity.

ahoiahoi:
@Rawaho good idea, thank you

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -145,6 +149,15 @@ public override void OnAddToMap(BaseMap map, uint guid, Vector3 vector)
             Session.EnqueueMessageEncrypted(new ServerPlayerEnteredWorld());
 
             IsLoading = false;
+
+            Position = vector;
+            Map.Entry.Id = (ushort)map.Entry.Id;
```
Rawaho: Both position and map are set before this in `base.OnAddToMap`, no need to do it again

Rawaho (changes requested):
Just a single small change

ahoiahoi:
@Rawaho updated

Rawaho (approved):

PR #66, opened 2018-12-09
Added ability to get/set properties via commands

jasoncouture:
This is useful for exploring the map.
Commands are:
/c set <property> <value>
and
/c set <property>

Property is parsed using Enum.TryParse, allowing for either enum name, or numeric value.

### DIFF
--- a/Source/NexusForever.WorldServer/Command/Handler/SetCommandHandler.cs
+++ b/Source/NexusForever.WorldServer/Command/Handler/SetCommandHandler.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Threading.Tasks;
+using NexusForever.WorldServer.Command.Attributes;
+using NexusForever.WorldServer.Command.Contexts;
+using NexusForever.WorldServer.Game.Entity.Static;
+
+namespace NexusForever.WorldServer.Command.Handler
+{
+    [Name("Game Master")]
+    public class SetCommandHandler : NamedCommand
+    {
+
+        public override string HelpText => "set <property> <value>";
+
+        public SetCommandHandler()
+            : base(true, "set")
+        {
+        }
+
+        protected override async Task HandleCommandAsync(CommandContext context, string command, string[] parameters)
+        {
+            if (parameters.Length != 2)
+            {
+                await SendHelpAsync(context);
+                return;
+            }
+
+            if (!Enum.TryParse(parameters[0], out Property property))
+            {
+                await context.SendErrorAsync($"Unknown property: {parameters[0]}").ConfigureAwait(false);
+                return;
+            }
+
+            if (!float.TryParse(parameters[1], out float propertyValue))
+            {
+                await context.SendErrorAsync($"Invalid property value: {parameters[1]}").ConfigureAwait(false);
+                return;
+            }
+
+            context.Session.Player.SetProperty(property, propertyValue);
+            context.Session.Player.TeleportTo((ushort) context.Session.Player.Map.Entry.Id,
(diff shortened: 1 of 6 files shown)


Source/NexusForever.WorldServer/Game/Entity/WorldEntity.cs
```diff
@@ -73,15 +73,15 @@ public virtual ServerEntityCreate BuildCreatePacket()
             };
         }
 
-        protected void SetProperty(Property property, float value, float baseValue = 0.0f)
+        public void SetProperty(Property property, float value, float baseValue = 0.0f)
```
Matt-Hurd, line 76: This method definitely shouldn't be public. I worry that making it public for a dev command will tempt people later on to use it in different contexts. I'd rather create a public function that is only available in debug.

Source/NexusForever.WorldServer/Command/Handler/SetCommandHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+            }
+
+            if (!float.TryParse(parameters[1], out float propertyValue))
+            {
+                await context.SendErrorAsync($"Invalid property value: {parameters[1]}").ConfigureAwait(false);
+                return;
+            }
+
+            context.Session.Player.SetProperty(property, propertyValue);
+            context.Session.Player.TeleportTo((ushort) context.Session.Player.Map.Entry.Id,
```
Rawaho, line 41: I would prefer this to use the correct packet and not an entire teleport to update

Source/NexusForever.WorldServer/Game/Entity/WorldEntity.cs
jasoncouture: There were official GM commands for this, and there will need to be a public way to achieve this.
Official commands I am referencing are /c runspeed and /c fly
I don't know the right way to go about it, but having this be only available in debug won't do when we do implement those commands.

Matt-Hurd: I'm not convinced that either of those commands need to be implemented by setting properties. I would expect there to be a buff or spell that is applied to modify these things. Considering there are normal buffs ingame for speed modification (explorer flags), as well as flying (shiphand when you're a robot, maybe others that I'm not remembering), this feels very wrong and I'd rather avoid giving the people implementing those features the excuse of "but it's public!".

jasoncouture: I agree with you. We're saying the same thing differently. I will leave this PR open and make what I think would be a sane way to do this, and then close it after I show you what I mean.

Source/NexusForever.WorldServer/Command/Handler/SetCommandHandler.cs
kirmmin: `0x093A` seems to be an opcode that will read properties. Not sure if it applies related properties to the client (I didn't dig into it too much) but could be a good place to start. Struct below:

```
ReadBits(32) // Possibly Player GUID - Needs to be > 0
const count = ReadBits(8)
while(count)
{
    Property = ReadBits(8)
    BaseValue = ReadBits(32)
    Value = ReadBits(32)
}
```

Edit 1: Fixed Struct because I incorrectly read it :)

jasoncouture:
Replaced by #82

Source/NexusForever.WorldServer/Command/Handler/SetCommandHandler.cs
jasoncouture: @kirmmin Thanks, fixed in #82

PR #70, opened 2018-12-12
Create Travis Config for Build and Migrations

Matt-Hurd:
This config will ensure that the project builds correctly, and that all the DB migrations work properly.

Currently using mono as the build matrix, but dotnet core is also available. 

Important note from https://docs.travis-ci.com/user/languages/csharp/#build-environment

> Note that these runtimes do not implement the entire .NET framework, so Windows .NET framework programs may not be fully compatible and require porting.

This sets up an environment that will allow for both unit and integration testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: csharp
+solution: Source/NexusForever.sln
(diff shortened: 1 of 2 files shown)


Rawaho:
I like this, the only issue I have with this is building with Mono and not .NET Core
The note from the docs is moot for us since we are targeting .NET Core

PR #73, opened 2018-12-18
Entitlement updates

kirmmin:
Pulled information from the Entitlement GameTable. There are a number of rewards for each one - some are titles, some are items, some are things like bag spaces.


kirmmin:
Looks like @Matt-Hurd's receive command got pulled in and removed in final PR. I don't want to bloat commit history so I'll close this one out and cherry pick the entitlement update

PR #81, opened 2018-12-30
Fixed issues preventing go command from working.

jasoncouture:
Issue #1: French was not loading correctly due to multi-byte unicode code points.
Issue #2: File cache was loading old data, added code to clean the cache if the binary changes.
Issue #3: Go command search was only returning a single ID, when there may be multiple.
Issue #4: *.bin has multiple ID's for a single string in some instances, the index was not accounting for this, which caused some random lookup failures.


jasoncouture:
Fixes #86

PR #84, opened 2019-01-01
Prevent duplicate account creation

kirmmin:


VoidEssy:
Maybe you want to target it for Develop?

kirmmin:
This hasn't been updated yet; It's on my list to do, but speaking with Rawaho, it needed to be handled by the Auth server, not in this way. I'll get around to updating it, but won't be just yet

PR #90, opened 2019-01-07
Abilities and basic casting

ahoiahoi:
* rebased and isolated from persistent stats/properties
* now with spell effects, in colour (tm)


ahoiahoi:
sigh, not so isolated -.-

PR #96, opened 2019-01-25
SRP6 fix

ahoiahoi:
this should fix one of the login issues
there is still an issue on the STS server when trying to re-login with a already authenticated session


Source/NexusForever.Shared/Cryptography/Srp6Provider.cs
```diff
@@ -160,7 +160,11 @@ private static BigInteger ShaInterleave(BigInteger key)
                 .Reverse()
                 .ToArray();
 
-            int length = Math.Max(4, Array.IndexOf<byte>(T, 0) + 1);
+            int first0 = Array.IndexOf<byte>(T.Reverse().ToArray(), 0);
```
Therzok: Why double reverse?

ahoiahoi: to not alter the code any more than needed

ahoiahoi:
Reopen

ahoiahoi:
reopened after takedown

PR #103, opened 2019-02-21
persistent stats

ahoiahoi:
* basic stat handling / client updates
* Persistent stats for players
* NPC stats loaded from DB, not saved for now

### DIFF
--- a/Database/Updates/Character/2019_02_19_stats.sql
+++ b/Database/Updates/Character/2019_02_19_stats.sql
@@ -1,0 +1,14 @@
+CREATE TABLE `character_stats` (
+    `id` BIGINT(20) UNSIGNED NOT NULL DEFAULT '0',
+    `stat` TINYINT UNSIGNED NOT NULL DEFAULT '0',
+    `value` FLOAT NOT NULL DEFAULT '0',
+    PRIMARY KEY (`id`, `stat`),
+    CONSTRAINT `FK__character_stats_stat_id_character_id` FOREIGN KEY (`id`) REFERENCES `character` (`id`) ON DELETE CASCADE
+);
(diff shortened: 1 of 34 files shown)


kirmmin:
Hey Ahoi,

I've been looking into this PR as I'm working on related things currently. Couple notes I have:

- I'm not positive, but it seems like we shouldn't be storing stats in the DB for characters - world entities makes sense, though. Seems like character stats should be calculated? We have all the modifiers to hand (theoretically) so seems like an unnecessary table. Would a world table for base stats be better?
    - There's some knock-on effects of this change, like moving `Level` to a stat and a different table. I'm not sure if it matters all that much, but I guess it's possible the DB table relationships become a little more complex than they need to be?
- I really like the idea of the `StatValue` Class "owning" the types. Good job!
- I'm working on the Xp packet & associated level up "trigger". I note that you have a `LevelUp()` method in this PR. I took a look at some other emulators and they seem to all have something a little more specific, like `GrantLevel(byte level)`. It's very nit-picky, and maybe this is where @Rawaho chimes in and tells me I'm wrong, but was just curious if there was a reason that we don't work in the same way?

Please don't take this as criticism, as I'm not intending to criticise. Just wanted to give you my thoughts and do my best to help get stat updating in as soon as possible, in a way that doesn't require huge modification later. It'll help a lot with some other packet work and give us a base for testing item modifiers and the like later on.

ahoiahoi:
Hi Kirmmin

Properties can be calculated (although the health value is off) from the tables.
most Stats are volatile but some require storage (e.g. health, level)

Stats can be accessed like any other property on the Entity, for even easier access one could create a GetLevel accessor.
The Level is a Stat, as such it should be stored. But that is just my opinion.
I don't see a reason why anything should require accessing the level by the DB, so it does not matter if it is in a separate table

The LevelUp is more of a demo, and I am very emotionless if it gets removed.

PR #106, opened 2019-03-02
Server launcher implementation

MaxtorCoder:
TODO: Move the 3 paths to 1 path and load the server DLLs via that.


Source/NexusForever.ServerLauncher/Form1.cs
```diff
@@ -0,0 +1,201 @@
...
+using NexusForever.ServerLauncher.Configuration;
+using System;
+using System.Diagnostics;
+using System.Drawing;
+using System.IO;
+using System.Windows.Forms;
+
+namespace NexusForever.ServerLauncher
+{
+    public partial class Form1 : Form
```
kirmmin, line 11: It might be better to have this Class named appropriately, so when other devs end up working on this in the future, it's a little clearer what this form is for :)

MaxtorCoder: Will do, thanks!

MarkWilds: I was wondering the servers are going to be windows only?

MaxtorCoder: I just made this for the sake of having a quick launcher. The server itself is made in .NET Core which is cross platform.

PR #112, opened 2019-03-04
Social: Whisper & Chat Channel Message Handling

kirmmin:
- Handles whisper messages between 2 clients.
- Handles Nexus & Trade chat channels and related messaging
- Handles formatting chat links.
- Abstracted chat link handling for use in other channel types if they use different opcodes, like whispers do.
- Added `Rules` object to WorldServer configuration. (Optional Parameter)
   - Added `CrossFactionChat` parameter to WorldServer configuration. To be used to enable/disable whispers cross faction. Defaults to Enabled. (Optional Parameter)

### DIFF
--- a/Source/NexusForever.Shared/Network/GameSession.cs
+++ b/Source/NexusForever.Shared/Network/GameSession.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -23,7 +22,7 @@ namespace NexusForever.Shared.Network
 
         private FragmentedBuffer onDeck;
         private readonly ConcurrentQueue<ClientGamePacket> incomingPackets = new ConcurrentQueue<ClientGamePacket>();
-        private readonly Queue<ServerGamePacket> outgoingPackets = new Queue<ServerGamePacket>();
+        private readonly ConcurrentQueue<ServerGamePacket> outgoingPackets = new ConcurrentQueue<ServerGamePacket>();
 
         /// <summary>
         /// Enqueue <see cref="IWritable"/> to be sent to the client.
--- a/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
+++ b/Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
@@ -94,6 +94,8 @@ namespace NexusForever.Shared.Network.Message
         ServerChatAccept                = 0x01C2,
         ClientChat                      = 0x01C3,
         ServerChat                      = 0x01C8,
+        ClientChatWhisper               = 0x01D4,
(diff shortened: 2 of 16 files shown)


Source/NexusForever.WorldServer/Network/Message/Model/Client018F.cs
```diff
@@ -0,0 +1,16 @@
+using NexusForever.Shared.Network;
+using NexusForever.Shared.Network.Message;
+
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    /// <summary>
+    /// Sent after client has processed logging in. This packet was appropriate to execute chat channel join amongst other things.
+    /// </summary>
```
Rawaho: This comment doesn't seem to be correct, 0x018F is sent from multiple client functions including from P2PTrading.CancelTrade

kirmmin: Gotcha. I just chose the last client opcode and it worked, I didn't dig further :P I will dig into the packets and find the right one. Thanks!

kirmmin: I haven't made any headway on this. I'm looking at these packets that are sent at login, but I'm not sure which one it could be. 
```
Client [0x03EC] Unknown
Client [0x03ED] Unknown
Client [0x0269] Unknown
Client [0x012B] Unknown
Client [0x03ED] Unknown
Client [0x07CC] Unknown
Client [0x07CC] Unknown
Client [0x07CC] Unknown
Client [0x07CC] Unknown
Client [0x07CC] Unknown
Client [0x07CC] Unknown
Client [0x07CC] Unknown
Client [0x018F] Unknown
```
Most of them are empty packets: `0x03EC`, `0x03ED`, `0x0269`, `0x018F`

If I make the Server wait until it receives `0x03EC`, it'll work fine. If I send it earlier, the client won't display the "You've joined X channel" message. Any insight would really help :)

Rawaho: Looking at what LUA functions call these opcodes will give more insight into what they do.

0x03EC is called by RequestOwnedCommodityOrders
0x03ED is called by RequestOwnedItemAuctions
0x0269 is called by RequestExchangeInfo
Not 100% sure on 0x012B but it is called in the same function with OpenOptions
0x07CC is called by RequestRewardUpdate, GetRewardRotations and a few others

There might not be a direct client message that requests the channels, you might just have to sent it in the SentPacketsAfterAddToMap function.

kirmmin: As long as I send it at the very end of `SendPacketsAfterAddToMap` method, it works just fine; Anything before that and it doesn't work.

This doesn't mimic Live, but it means I'm not using a Client packet that's unnecessary, so I think that's fine.

PR #114, opened 2019-03-12
Weapon Sheathed Toggle

kirmmin:
Replaces #85 

- Sheath functionality as preview of this feature.


Source/NexusForever.WorldServer/Network/Message/Model/ServerStatUpdateFloat.cs
```diff
@@ -0,0 +1,23 @@
...
+using NexusForever.WorldServer.Game.Entity.Static;
+
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    [Message(GameMessageOpcode.ServerStatUpdateFloat, MessageDirection.Server)]
+    public class ServerStatUpdateFloat : IWritable
+    {
+        public uint Guid { get; set; }
+        public Stat Stat { get; set; }
+        public uint Value { get; set; }
```
Therzok: Shouldn't this be float?

kirmmin: Maybe. It's been a few months since I originally put this together so thanks for the comment - I will get this tested & updated.

MaxtorCoder: > 
> 
> Shouldn't this be float?

There are 2. Float and int. Maybe even more.

kirmmin:
This is not ready for merge. After #103 gets merged, then I will update this PR to reflect changes made.

PR #119, opened 2019-03-19
ZoneID Blacklist

badbaki:
Setup an enum with all currently known bad zoneIDs and implemented a check on the Player.cs TeleportTo function to prevent travel and notify the user that the zoneID is blacklisted.

Zones that have yet to be implemented, like housing communities, were not added. This is just the zoneIDs known to fatally crash the world server.


Ness199X:
You need to rebase your commit on the most recent commit to develop

badbaki:
Rebased branch to latest develop commit.

badbaki:
Updated to current development build

Anaherra:
The information from this pull request has been added the new disable manager in e73a5c8e74b8a7f5b62ddd7436925de8ced7a021
Your research into which worlds were causing issues was very helpful, thank you for the contribution!

Issue #25, opened 2018-11-30
Social Tasks


n30na:
- [ ] Chat
- [ ] Friends list
- [ ] Party & raid formation
- [ ] Party chat
- [ ] Housing
- [ ] Guild formation and player list
- [ ] Other basic guild functionality (motd etc)
- [ ] Circles
- [ ] Guild exp/perks
- [ ] Guild bank

lucas-jones:
Say, Yell & Emote chat is done :)

PR #120, opened 2019-03-20
Social/Contacts: Contacts Feature

kirmmin:
**Features**
- _Contact Manager_
   - Centralised contact manager to handle manipulation of contacts.
   - Caches all contacts to ensure whether players are online or offline, the contact can be still be stored and delivered when that user logs on.
   - Minimal player code
- _Character Friends_
   - Friend requests can be sent between players
   - Friend requests can be accepted, declined, and ignored.
   - Notifications should be sent when a friend logs on or off to all friends currently online.
      - "Last online" is working, but not implemented fully yet.
   - Friend icon shows for nearby friends.
   - Removal of Friends 
        - Does not remove the player initiating removal from the friend if they are a friend. Optimisation that can be added later on.
   - Transfer from friend to rival or ignore.
- _Rivals_
   - Rivals can be added and removed.
   - Rival icon shows for nearby rivals.
- _Ignored_
   - Players can be added and removed from ignore.
   - Rudimentary blocking of ignored users from nearby chat channels (say, yell, emote).
      - Super basic and likely to be modified/removed when session tracking is overhauled.
- _Player Information_
   - When the client asks for information on a player, it regularly uses the `0x0597` packet to request it. This has been implemented as part of this feature and is also being used for Mail (at minimum).
- _TBD_
   - Account Friends - Packets figured out, but not implemented
   - Neighbours

**Commands**
- None! All features utilise the UI itself.

Should merge after #125

Contributes to #25


Source/NexusForever.Shared/Network/GamePacketWriter.cs
```diff
@@ -129,5 +129,14 @@ public void WriteStringWide(string value)
             Write(data.Length >> 1, extended ? 15u : 7u);
             WriteBytes(data);
         }
+
+        public void WriteStringFixed(string value)
+        {
+            string str = $"{value.Replace("\0", "") ?? ""}\0";
```
Fabi: `value` can be null here. You want to use the `??` check on `value` or use `value?.Replace`

Source/NexusForever.WorldServer/Game/Contact/ContactManager.cs
```diff
@@ -0,0 +1,774 @@
...
+        /// <summary>
+        /// Accept a <see cref="Contact"/> request
+        /// </summary>
+        /// <param name="session">Session of the player accepting the request</param>
+        /// <param name="contactId">Contact.ID of the accepted request</param>
+        /// <param name="returnRequest">True if the player is making the requester a friend as well</param>
+        public static void AcceptFriendRequest(WorldSession session, ulong contactId, bool returnRequest = false)
+        {
+            contactsCache.TryGetValue(contactId, out Contact contactRequest);
+            if (contactRequest == null)
```
Fabi: I think you can use `if (!contactsCache.TryGetValue(contactId, out Contact contactRequest))` here instead of that additional null check?

Also throwing an `Exception` without any other info specific is not that nice.

Source/NexusForever.WorldServer/Game/Contact/ContactManager.cs
```diff
@@ -0,0 +1,774 @@
...
+        /// <summary>
+        /// Decline a <see cref="Contact"/> request
+        /// </summary>
+        /// <param name="session">Session of the player declining the request</param>
+        /// <param name="contactId">Contact.ID of the declined request</param>
+        /// <param name="addIgnore">Flag to create an ignore entry for this user</param>
+        public static void DeclineFriendRequest(WorldSession session, ulong contactId, bool addIgnore = false)
+        {
+            contactsCache.TryGetValue(contactId, out Contact contactRequest);
+            if (contactRequest == null)
```
Fabi: Look up

Source/NexusForever.WorldServer/Game/Contact/ContactManager.cs
```diff
@@ -0,0 +1,774 @@
...
+        /// <summary>
+        /// Delete <see cref="Contact"/> from a player's contacts
+        /// </summary>
+        /// <param name="session">Session of player making the deletion</param>
+        /// <param name="characterIdentity"><see cref="CharacterIdentity"/> of the player to delete</param>
+        /// <param name="type">Type to confirm deletion with</param>
+        public static void DeleteContact(WorldSession session, CharacterIdentity characterIdentity, ContactType type)
+        {
+            Contact contactToDelete = contactsCache.Values.FirstOrDefault(s => s.OwnerId == session.Player.CharacterId && s.ContactId == characterIdentity.CharacterId && !s.IsPendingDelete);
+            if (contactToDelete == null)
```
Fabi: Same here with throwing a simple `Exception`. Not sure if throwing at these things is intended at all or if some kind of logging is enough. If throwing is needed specify the exception a bit more^^

Source/NexusForever.WorldServer/Game/Contact/ContactManager.cs
```diff
@@ -0,0 +1,774 @@
...
+        /// <summary>
+        /// Set a private note associated with a <see cref="Contact"/>
+        /// </summary>
+        /// <param name="session">Sesion of the player making the note change</param>
+        /// <param name="characterIdentity">CharacterIdentity of the player's note to update</param>
+        /// <param name="note">Note to set</param>
+        public static void SetPrivateNote(WorldSession session, CharacterIdentity characterIdentity, string note)
+        {
+            Contact contactToModify = contactsCache.Values.FirstOrDefault(s => s.OwnerId == session.Player.CharacterId && s.ContactId == characterIdentity.CharacterId && !s.IsPendingDelete);
+            if (contactToModify == null)
```
Fabi: Look up

kirmmin:
Rebased. I messed up the commit messages, and I need to fix a few things after the rebase which will happen tomorrow.

kirmmin:
Refactored this code in entirety. This was a bigger task than I had anticipated, but has a much cleaner implementation now.

I've also fixed some minor bugs that the Contacts system relied on, namely the Level of the Player when loaded by the `CharacterManager` from the database. And, the `lastOnline` not saving (EF Core doesn't update date fields on save, automatically, you must tell it to do so, even with the `ValueGeneratedOnAddOrUpdate()` method applied to the `entity`).

This now uses a subscription method, and the `WorldSession` passing has been removed entirely. It only passes Character IDs and we lookup sessions from the Global scope. The rest of the time, it has direct access to the Session via the Contact Manager's owner (a `Player` instance).

I have also refactored this to use DB Migrations. Believe I am getting the hang of the migrations, now, and beginning to really enjoy the iterative process :) So much easier than modifying SQL then the code itself to match. I modify once, run a couple commands, and can test out a new data set 👍 Really appreciate this design pattern.

Finally, I have removed WIP. I have no plans to add more to this PR, and it's massive already.

Source/NexusForever.WorldServer/Game/Contact/ContactRules.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            Dictionary<ContactType, uint> maxTypeMap = new Dictionary<ContactType, uint>
+            {
+                { ContactType.Friend, GetMaxFriends() },
+                { ContactType.Account, GetMaxFriends() },
+                { ContactType.Ignore, GetMaxIgnored() },
+                { ContactType.Rival, GetMaxRivals() }
+            };
+            Dictionary<ContactType, ContactResult> maxTypeResponseMap = new Dictionary<ContactType, ContactResult>
+                {
```
MaxtorCoder: Formatting

kirmmin: Thanks Max ;)

kirmmin:
Rebased to `game_rework` branch. Certainly a bit involved, but I like the cleanliness of the `interfaces` :) Not ready for merge yet, as I have to go over the database migration and compare it to test server to make sure I didn't make any further changes, but I plan on doing that tomorrow.

kirmmin:
Checked and recreated database migration. Also includes a fix for Guild Member login/logout functionality. This is reliant on some changes I made as part of this PR to work appropriately now that we have `PlayerManager` and `CharacterManager`.

I'm not going to expand functionality further as part of this PR. After review and, hopefully, mergng, I can visit some of the changes. It's already too massive :)

Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
```diff
@@ -24,14 +28,15 @@ public static void HandlePing(IWorldSession session, ClientPing ping)
         [MessageHandler(GameMessageOpcode.ClientPlayerInfoRequest)]
         public static void HandlePlayerInfoRequest(IWorldSession session, ClientPlayerInfoRequest request)
         {
-            ICharacter character = CharacterManager.Instance.GetCharacter(request.Identity.CharacterId);
+            ICharacter character = PlayerManager.Instance.GetPlayer(request.Identity.CharacterId) as Player;
```
kirmmin: @Rawaho I've had to do this "double-check" in a few places now. `PlayerManager` returns a "live" player and `CharacterManager` returns a cached version. 
As part of the rework, it appears as though you separated `CharacterManager` into 2, which seemed necessary, given that we needed a new way to get Players and their sessions. The knock-on effect is it makes these things a little messy. I suppose I could make like a `PlayerHelper` or something to abstract this check? I assume we'll need something like this in other places.

The only alternative I can see is `PlayerManager` updating the `CharacterManager` data semi-frequently, but that seems bad?

kirmmin: I thought about it for 10 seconds more :) How would you feel if `CharacterManager` called out to `PlayerManager` when it was requested data, it would then update its own objects and then ship out the data to the requesting function?
 
That would make things a lot cleaner, CharacterManager would always be the "I need data about the player, regardless of online status", and we only need to call out to PlayerManager if we need an "online" Player to work with.

kirmmin: @Rawaho Please check out the latest commit (https://github.com/NexusForever/NexusForever/pull/120/commits/96bdbde3162378ee4e5b55a3ec766fbf35f97cbd) to see what I was referring to with my above comment, if I didn't explain well :) I separated out the commit for ease of viewing, but will squash if you're happy.

This keeps the managers' purposes unique without complicating lookups.
- `PlayerManager` is the way to get a logged in `IPlayer` instance.
- `CharacterManager` is used to get "static" character information when you don't need an `IPlayer` instance.

Rawaho:
After 2,602 days this PR can finally be closed, better late than never right?
Was refactored as part of commit 99752578ec9bd575b96f7d60f85ca10dc836eb15

PR #124, opened 2019-03-24
CurrencyManager - Fix CanAfford method

kirmmin:


Source/NexusForever.WorldServer/Game/Entity/CurrencyManager.cs
```diff
@@ -152,7 +152,7 @@ public bool CanAfford(byte currencyId, ulong amount)
             if (!currencies.TryGetValue(currencyId, out Currency currency))
                 return false;
 
-            return currency.Amount < amount;
+            return currency.Amount > amount;
```
Rawaho: Shouldn't this be >= ?

Fabi: > Shouldn't this be >= ?

yes

kirmmin: Oh, yea, lol. It was late and didn't trigger, thanks. I'll get that updated today

PR #125, opened 2019-03-24
Player Info: Delivering player info

kirmmin:
- Used by Mail & Contacts, and probably other features to update the client about Names, Class, Race, etc. for associated Character ID
- Also includes the `WriteStirngFixed` method for `GamePacketWriter`

TODO: Probably needs some optimisation, like caching player data and not needing to make a DB request each time.

### DIFF
--- a/Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
+++ b/Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
@@ -1,10 +1,11 @@
 using NexusForever.Shared.Game.Events;
+using NexusForever.Shared.Network;
 using NexusForever.Shared.Network.Message;
 using NexusForever.WorldServer.Database.Character;
 using NexusForever.WorldServer.Database.Character.Model;
 using NexusForever.WorldServer.Game.Entity.Static;
 using NexusForever.WorldServer.Network.Message.Model;
-using System.Threading.Tasks;
+using NexusForever.WorldServer.Network.Message.Model.Shared;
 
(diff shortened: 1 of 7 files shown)


Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
```diff
@@ -10,5 +13,30 @@ public static void HandlePing(WorldSession session, ClientPing ping)
...
+        /// <summary>
+        /// Handled responses to Player Info Requests.
+        /// TODO: Put this in the right place, this is used by Mail & Contacts, at minimum. Probably used by Guilds, Circles, etc. too.
+        /// </summary>
+        /// <param name="session"></param>
+        /// <param name="request"></param>
+        [MessageHandler(GameMessageOpcode.ClientPlayerInfoRequest)]
+        public static void HandlePlayerInfoRequest(WorldSession session, ClientPlayerInfoRequest request)
+        {
+            Character character = CharacterDatabase.GetCharacterById(request.CharacterId);
```
Rawaho: This should be asynchronous and handled as an event, having a blocking query in a session is bad due to the fact that all sessions are updated sequentially.

Rawaho (changes requested):

Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
kirmmin: This should be resolved, now. Pretty simple change, too. I like the simplicity of this :)

Rawaho (approved):

PR #137, opened 2019-04-02
Inventory: Bag support, Error support, Bug fixes

kirmmin:
**Features**
- Add, Remove Bags from Inventory
- Add, Remove Items from Bags
- Should not be able to move or swap a bag that contains items
- Item Error packet added to handle numerous item conditions

~Requires #122 to be merged~


Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -382,6 +391,24 @@ public void ItemMove(ItemLocation from, ItemLocation to)
...
+
+        /// <summary>
+        /// Move <see cref="Item"/> from one <see cref="ItemLocation"/> to another, handling swapping if necessary.
+        /// </summary>
+        private void ItemMove(ItemLocation from, ItemLocation to, Item srcItem, Item dstItem = null)
+        {
+            if (srcItem == null || from == null || to == null)
+                throw new InvalidPacketValueException();
+
+            if (dstItem == null && GetBag(to.Location).GetItem(to.BagIndex) != null)
```
MarbleBag: This overloaded version of ItemMove does validate its arguments against null, but the other doesn't. 
GetBag(to.Location) can return null

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -382,6 +391,24 @@ public void ItemMove(ItemLocation from, ItemLocation to)
...
+            if (IsEquippedBagLocation(from) || IsEquippedBagLocation(to))
+                HandleEquippedBagChange(from, to, srcItem, dstItem);
+            else
+                ItemMove(from, to, srcItem, dstItem);
+        }
+
+        /// <summary>
+        /// Move <see cref="Item"/> from one <see cref="ItemLocation"/> to another, handling swapping if necessary.
+        /// </summary>
+        private void ItemMove(ItemLocation from, ItemLocation to, Item srcItem, Item dstItem = null)
```
MarbleBag: dstItem is not a descriptive name to indicate, that this argument is required for swapping.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -605,6 +634,138 @@ private void ItemStackCountUpdate(Item item, uint stackCount)
...
+                else if (dstItem != null && !IsEquippedBagLocation(from))
+                    capacityChange = (int)srcItem.Entry.MaxStackCount - (int)dstItem.Entry.MaxStackCount; // Replacing bag
+            }
+            else if (srcItem.IsEquippableBag() && IsEquippedBagLocation(from))
+            {
+                if (dstItem == null && !IsEquippedBagLocation(to))
+                    capacityChange -= (int)srcItem.Entry.MaxStackCount; // Removing bag (-1 for actual bag)
+                else if (dstItem != null && dstItem.IsEquippableBag())
+                    capacityChange = (int)dstItem.Entry.MaxStackCount - (int)srcItem.Entry.MaxStackCount; // Replacing bag
+            }
```
MarbleBag: What does the client do, if the player swaps two equipped bags? Does it send two ItemMoves?
If not, what happens if the player swaps two equipped bags which have a different size?

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -605,6 +634,138 @@ private void ItemStackCountUpdate(Item item, uint stackCount)
...
+
+            if (capacityChange < 0 && inventory.SlotsRemaining < (capacityChange * -1 + 1))
+                itemError = ItemError.BagMustBeEmpty;
```
MarbleBag: Why is it not possible to swap a 14 slot bag, with a 10 slot bag, if 4 inventory slots are empty?
capacityChange  := -4
4 = inventory.SlotsRemaining < (capacityChange * -1 + 1) = (-4*-1+1) = 5
Why is that an error?

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -605,6 +634,138 @@ private void ItemStackCountUpdate(Item item, uint stackCount)
...
+                itemError = ItemError.InvalidForThisSlot;
+
+            // TODO: Mail items it can't bag, instead of erroring? Or, flow into overflow?
+            int capacityChange = 0;
+            if (srcItem.IsEquippableBag() && IsEquippedBagLocation(to))
+            {
+                if (dstItem == null && !IsEquippedBagLocation(from))
+                    capacityChange = (int)srcItem.Entry.MaxStackCount; // Adding bag
+                else if (dstItem != null && !IsEquippedBagLocation(from))
+                    capacityChange = (int)srcItem.Entry.MaxStackCount - (int)dstItem.Entry.MaxStackCount; // Replacing bag
```
MarbleBag: Why does nothing happen, if both locations (to & from) return true on IsEquippedBagLocation?

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: If the player swaps two equipped bag, it sends one move request. No calculations need to happen, and the item positions are just swapped.
Same thing happens when 2 bags are swapped which have different sizes. If they are both equipped, there's no need to consider resizing as the bags are only changing position in the same context.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: There are certain occasions, like the one you mentioned, where the +1 on. the end messes with it. It needs to be cleaner but requires digging through it all. The +1 is only really required when you're removing a bag, but I need to figure out a way to toggle that appropriately.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: See comment above in response to two equipped bags.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: I kept the naming convention that was in place. The same processing should be in place as everything needs to go through `ItemMove(from, to)` first, to get to this override, given that it's private to the Inventory Class.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: This line could technically be removed. I originally inserted it to ensure that, when moving items, we weren't trying to place the item in a location that was already being used. With that said, I refactored the process of removing a bag and moving items, and this line no longer triggered, and likely won't, now. It should only fire if other item manipulation occurs before the requested item move happens - should be impossible.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
VoidEssy: I really have issues with capacityChange *-1 + 1 feels like a hack 
Lets assume capacityChange(20)
capacityChange * -1 = -capacityChange (-20)
-capacityChange +1 = -capacityChange(-19)

maybe just maybe (capacityChange - 1) * -1 (But still a bit of a WTF for me.) I assume it's because Arrays still probably there is a better way about it.

Alternatively
```
 if (capacityChange < 0 && inventory.SlotsRemaining < --capacityChange * -1)
```
or
```
 if (capacityChange < 0 && inventory.SlotsRemaining < capacityChange-- * -1)
```
Was lazy to start-up proj just to verify used Fiddle which apparently doesn't react to capacityChange-- but does to --capacityChange but both should be interchangeable


Okay brushed up --x and x-- is pre and post decrement just never had the need to use pre decrement. Both should be valid in this case.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
VoidEssy: If dstItem == null shouldn't it just do nothing? Instead of executing the code? I'd remove Item dstItem = null and add a check 
```
private void ItemMove(ItemLocation from, ItemLocation to, Item srcItem, Item dstItem)
if(dstItem != null)
{
DoSomething
}
```
else it just skips out. I think faster that way?

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: if `capacityChange` was 20, then this line wouldn't execute. The value must be negative to process.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: If `dstItem` is `null`, then we don't need to worry about moving the item that exists in the position we're moving an item to. We still need to process the move request, though, we can't do nothing.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: I've pushed an update that should be a little clearer to read. It's not a complex calculation, but it's very difficult to simplify because there's a few scenarios being distilled into 1 check.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
VoidEssy: Okay yeah kinda makes sense but again maybe it's just my lack of knowledge of deeper intricacies if you can't not do anything then why not have default response for no-action? if that's what it ties to?

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
VoidEssy: Yeah I understand that the value has to be negative hence the * -1 but i guess you could do a clean-up with the pre decrementation as shown in the example above. Sorry if you didn't see it i was editing the response.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
kirmmin: If there is no action to be completed, we shouldn't get to this override. It should be blocked in the public `ItemMove` method.

VoidEssy: Alright then I've nothing else to add. Thanks for clearing up.

Source/NexusForever.WorldServer/Game/Entity/Bag.cs
```diff
@@ -98,12 +112,11 @@ public void Resize(int capacityChange)
         {
             if (items.Length + capacityChange < 0)
                 throw new ArgumentOutOfRangeException(nameof(capacityChange));
-            if (items.Length + capacityChange < items.Length)
-                throw new ArgumentOutOfRangeException(nameof(capacityChange));
 
             Array.Resize(ref items, items.Length + capacityChange);
+            SlotsRemaining = items.Length - items.Count(i => i != null);
```
Therzok: I don't think we need to Count the items here. We can just add `capacityChange` to SlotsRemaining.

```suggestion
SlotsRemaining += capacityChange;
```

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -732,6 +767,158 @@ public bool ItemUse(Item item)
             return true;
         }
 
+        /// Check if the current <see cref="Item"/> is in a visual slot
+        /// </summary>
+        private bool IsVisualItem(EquippedItem item)
+        {
+            EquippedItem[] visualItems = new EquippedItem[]
+            {EquippedItem.Chest, EquippedItem.Head, EquippedItem.Legs, EquippedItem.Hands, EquippedItem.WeaponPrimary, EquippedItem.Shoulder, EquippedItem.Feet};
```
Therzok: Maybe cache this to avoid allocating an array every time?

Therzok:
```
static EquippedItem[] visualItems = new EquippedItem[] { EquippedItem.Chest, EquippedItem.Head, EquippedItem.Legs, EquippedItem.Hands, EquippedItem.WeaponPrimary, EquippedItem.Shoulder, EquippedItem.Feet };

File: Source/NexusForever.WorldServer/Game/Entity/Inventory.cs [hunk 8]
```diff
@@ -732,6 +767,158 @@ public bool ItemUse(Item item)
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Check if the <see cref="Item"/> is an equipped container
+        /// </summary>
+        private bool IsEquippedBag(Item item)
+        {
+            EquippedItem[] containers = new EquippedItem[]
```
Therzok: Same here.

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -732,6 +767,158 @@ public bool ItemUse(Item item)
...
+            return IsEquippedBagLocation(itemLocation.Location, itemLocation.BagIndex);
+        }
+
+        /// <summary>
+        /// Returns whether or not the provided <see cref="InventoryLocation"/> and bagIndex match an Equipped Bag slot.
+        /// </summary>
+        private bool IsEquippedBagLocation(InventoryLocation inventoryLocation, uint bagIndex)
+        {
+            EquippedItem[] containers = new EquippedItem[]
+            {EquippedItem.Bag0, EquippedItem.Bag1, EquippedItem.Bag2, EquippedItem.Bag3};
```
Therzok: Cache.

kirmmin:
Updated following review comments. Also found a minor bug where the slots weren't initially set if the Player did not have any bags already equipped.

This code has been on my test server for a couple months. A couple bugs were found and resolved, so I believe this is ready for merge.

kirmmin:
Rebased today with latest changes to `develop`.

kirmmin:
Rebased and updated today.

PR #139, opened 2019-04-06
Commands/Items: Lookup command

kirmmin:
- Item creation by item link
- ChatLinks formatted and passed to Command Handlers

### DIFF
--- a/Source/NexusForever.Database.Auth/AuthContext.cs
+++ b/Source/NexusForever.Database.Auth/AuthContext.cs
@@ -803,6 +803,11 @@ namespace NexusForever.Database.Auth
                     },
                     new PermissionModel
                     {
+                        Id   = 92,
+                        Name = "Command: ItemLookup"
+                    },
+                    new PermissionModel
+                    {
                         Id   = 10000,
                         Name = "Other: InstantLogout"
                     });
--- a/Source/NexusForever.Database.Auth/Migrations/20200522025519_ItemLookupPermission.Designer.cs
+++ b/Source/NexusForever.Database.Auth/Migrations/20200522025519_ItemLookupPermission.Designer.cs
@@ -2,15 +2,17 @@
 using System;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 using NexusForever.Database.Auth;
 
 namespace NexusForever.Database.Auth.Migrations
 {
     [DbContext(typeof(AuthContext))]
-    partial class AuthContextModelSnapshot : ModelSnapshot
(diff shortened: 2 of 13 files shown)


Rawaho:
I'm not really a fan of the way item links are being handled for item creation, the links should really be parsed into `IChatFormat` and passed to the handler in some way, the current method makes it seem quite hacky.

kirmmin:
> I'm not really a fan of the way item links are being handled for item creation, the links should really be parsed into `IChatFormat` and passed to the handler in some way, the current method makes it seem quite hacky.

This is on my list to do. I've been bouncing around my other PRs fixing up bugs I'm running into on the TestServer - doing my best to stay on top of the simple ones. I'll try and get this updated this week.

kirmmin:
Updated to include passing `IEnumerable<ChatFormat>` directly to Command Handlers for them to act on.

Source/NexusForever.WorldServer/Command/CommandManager.cs
```diff
@@ -31,17 +32,17 @@ public static IEnumerable<ICommandHandler> GetCommandHandlers()
             return commandHandlers.OrderBy(i => i.Order);
         }
 
-        public static bool HandleCommand(WorldSession session, string commandText, bool isFromChat)
+        public static bool HandleCommand(WorldSession session, string commandText, bool isFromChat, IEnumerable<ChatFormat> chatLinks = null)
         {
-            return HandleCommand(new WorldSessionCommandContext(session), commandText, isFromChat);
+            return HandleCommand(new WorldSessionCommandContext(session), commandText, isFromChat, chatLinks);
```
kirmmin: Was looking into this today, after working on command stuff over the past week (#155), I think it might be better to augment the `CommandContext` class to include a list of `ChatFormat`.

There's a couple reasons I think this is a better idea:
- Command signatures won't grow if we intend to add more variables in the future
- I want to be able to add a target Entity for GM commands - for example, turning a player into a different creature model temporarily, or updating a player's quest
- Eventually, we may want to automatically parse all the variables to reduce the command complexity and adding them all into `CommandContext` and keeping the signature small for further improvements later might really be a good option.
- Finally, setting the precedent that the signature shouldn't change is probably most important. Updating all these handlers, in the future, could be a nightmare, if we "approved" signature changing.

kirmmin: Refactored code in line with previous comment. Went from 20+ changed files to 6 and is far more scalable.

kirmmin:
Renamed command to lookup to stop triggering Rawaho ;)
Made some minor updates to item add to give feedback that the item was summoned.
Updated stack count update method to receive a reason, too.

kirmmin:
Rebased today to the latest `develop` branch.

kirmmin:
Rebased and updated today.

kirmmin:
Rebased this and refactored to the new command system. I have removed accepted chat links for now, but it is possible we may want to add that later.

PR #142, opened 2019-04-10
Character List location fix

ahoiahoi:
- set WorldZoneId correctly
- set MaxLevelAchieved correctly

### DIFF
--- a/Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
+++ b/Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
@@ -9,6 +9,7 @@ using NexusForever.Shared.Game;
 using NexusForever.Shared.Game.Events;
 using NexusForever.Shared.GameTable;
 using NexusForever.Shared.GameTable.Model;
+using NexusForever.Shared.IO.Map;
 using NexusForever.Shared.Network;
 using NexusForever.Shared.Network.Message;
 using NexusForever.WorldServer.Database.Character;
@@ -101,6 +102,9 @@ namespace NexusForever.WorldServer.Network.Message.Handler
             session.EnqueueEvent(new TaskGenericEvent<List<Character>>(CharacterDatabase.GetCharacters(session.Account.Id),
                 characters =>
             {
+                byte MaxCharacterLevelAchieved = 1;
+                var mapFiles = new Dictionary<ushort, MapFile>();
+
                 session.Characters.Clear();
                 session.Characters.AddRange(characters);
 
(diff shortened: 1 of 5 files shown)


Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
```diff
@@ -153,11 +153,22 @@ public static void HandleCharacterList(WorldSession session, ClientCharacterList
...
                         Path        = (byte)character.ActivePath
                     };
 
+                    if (character.Level > MaxCharacterLevelAchieved)
+                        MaxCharacterLevelAchieved = (byte)character.Level;
+
+                    if (!mapFiles.TryGetValue(character.WorldId, out MapFile mapFile))
+                    {
+                        WorldEntry entry = GameTableManager.World.GetEntry(character.WorldId);
+                        mapFile = BaseMap.LoadMapFile(entry.AssetPath);
```
Rawaho: I don't believe loading the BaseMap file every time a player is at the character screen is a good idea.
My suggestion would be to just store the zone in the database.

Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
```diff
@@ -153,11 +153,22 @@ public static void HandleCharacterList(WorldSession session, ClientCharacterList
                         Faction     = character.FactionId,
                         Level       = character.Level,
                         WorldId     = character.WorldId,
-                        WorldZoneId = 5967,
                         RealmId     = WorldServer.RealmId,
                         Path        = (byte)character.ActivePath
                     };
 
+                    if (character.Level > MaxCharacterLevelAchieved)
+                        MaxCharacterLevelAchieved = (byte)character.Level;
```
Rawaho: Math.Max would be cleaner here.

Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
ahoiahoi: how about we store the WorldZoneId instead of the WorldId?
The WorldId can be derived from WorldZoneId and there are inconsistencies already where the WorldId is not set to the corresponding WorldZoneId (e.g. walk/swim from Deradune to Crimson Isle)

Rawaho: I don't believe you can get the WorldId from the WorldZoneId, the same zones can appear on multiple worlds.
Unless I am missing something?

ahoiahoi: you are correct, I did not know that it was a N:N relation - I'll add the extra field then

PR #143, opened 2019-04-12
format fix

ahoiahoi:
this still pop's up as modified on a fresh checkout


kirmmin:
I've been fresh checking out for weeks now and not had this issue. Are you sure it's not just a bug on your end?

ahoiahoi:
I was surprised too as I didn't either - I'll check again

ahoiahoi:
so, it's confirmed - please proceed with merge :)

PR #145, opened 2019-04-22
Account: Add Basic Validation & Prevent Duplicates

kirmmin:
- Prevents creating accounts with special characters not allowed in emails.
- Prevents creating accounts with same email as another account
- Forces email/usernames to lowercase. The client sends emails/usernames to the server in lowercase, so this should prevent many login issues for new users or servers.

### DIFF
--- a/Source/NexusForever.Database.Auth/AuthDatabase.cs
+++ b/Source/NexusForever.Database.Auth/AuthDatabase.cs
@@ -76,8 +76,6 @@ namespace NexusForever.Database.Auth
                 .Include(a => a.AccountEntitlement)
                 .Include(a => a.AccountPermission)
                 .Include(a => a.AccountRole)
-                .Include(a => a.AccountPermission)
-                .Include(a => a.AccountRole)
                 .SingleOrDefaultAsync(a => a.Email == email && a.SessionKey == sessionKey);
         }
 


Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
```diff
@@ -50,27 +51,37 @@ public static async Task<Account> GetAccountAsync(string email, byte[] sessionKe
...
+                    return null;
+
+                // Ensure only Alphanumeric, '@', and '.', symbols are used in the email
+                if (Regex.IsMatch(email, @"[^A-Za-z0-9@.]"))
+                    return null;
+
                 byte[] s = RandomProvider.GetBytes(16u);
                 byte[] v = Srp6Provider.GenerateVerifier(s, email, password);
 
                 context.Account.Add(new Account
```
Therzok:
```suggestion
                var account = context.Account.Add(new Account
```

Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
```diff
@@ -50,27 +51,37 @@ public static async Task<Account> GetAccountAsync(string email, byte[] sessionKe
...
                 {
-                    Email = email,
+                    Email = email.ToLower(),
                     S     = s.ToHexString(),
                     V     = v.ToHexString()
                 });
 
                 context.SaveChanges();
+
+                return await context.Account.FirstOrDefaultAsync(a => a.Email == email.ToLower());
```
Therzok:
```suggestion
                return account;
```

Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
```diff
@@ -50,27 +51,37 @@ public static async Task<Account> GetAccountAsync(string email, byte[] sessionKe
...
         /// <summary>
         /// Create a new account with the supplied email and password, the password will have a verifier generated that is inserted into the database.
         /// </summary>
-        public static void CreateAccount(string email, string password)
+        public static async Task<Account> CreateAccount(string email, string password)
         {
             using (var context = new AuthContext())
             {
+                // Ensure account doesn't already exist with the same email
+                if (context.Account.FirstOrDefault(a => a.Email == email.ToLower()) != null)
```
Therzok: This will allocate a new string for `email.ToLower()` on every iteration of the loop.

Maybe just do `email = email.ToLower()` at the beginning of the method.

Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
```diff
@@ -82,7 +93,7 @@ public static bool DeleteAccount(string email)
             // Thanks Rawaho!
             using (var context = new AuthContext())
             {
-                Account account = context.Account.SingleOrDefault(a => a.Email == email);
+                Account account = context.Account.SingleOrDefault(a => a.Email == email.ToLower());
```
Therzok: Same problem as above, we're allocating a new string on every iteration of the loop.

Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
```diff
@@ -50,27 +51,37 @@ public static async Task<Account> GetAccountAsync(string email, byte[] sessionKe
                     .Include(a => a.AccountCostumeUnlock)
                     .Include(a => a.AccountGenericUnlock)
                     .Include(a => a.AccountKeybinding)
-                    .SingleOrDefaultAsync(a => a.Email == email && a.SessionKey == sessionKey);
+                    .SingleOrDefaultAsync(a => a.Email == email.ToLower() && a.SessionKey == sessionKey);
```
Therzok: Same allocation issue.

Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
Fabi: `.ToLower` should happen before this method is called

Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
CT-DeanSmith: I avoided doing this so I could confirm it had actually saved and there was no errors at the DB.

Fabi: Returning the account object would not guarantee that it's saved to the database.

kirmmin: I avoided doing this so I could confirm it had actually saved and there was no errors at the DB.

kirmmin:
Updated use of `ToLower()` to only be used a single time per method.

Therzok (approved):

Source/NexusForever.Shared/Database/Auth/AuthDatabase.cs
```diff
@@ -56,10 +60,19 @@ public static async Task<Account> GetAccountAsync(string email, byte[] sessionKe
...
         /// Create a new account with the supplied email and password, the password will have a verifier generated that is inserted into the database.
         /// </summary>
-        public static void CreateAccount(string email, string password)
+        public static async Task<Account> CreateAccount(string email, string password)
         {
+            email = email.ToLower();
             using (var context = new AuthContext())
             {
+                // Ensure account doesn't already exist with the same email
+                if (context.Account.FirstOrDefault(a => a.Email == email) != null)
```
Therzok: Should this use the async version too, now?

VoidEssy: I'd say no because it's not a fire-and-forget or return when ready. Synchronous functions that are part of a logical chain shouldn't be async unless the whole action is async in which case the initial method call should be async and everything down the line will be allocated to the executing thread.

EDIT: Async is fun and powerful but shoving it every-where is recipe to disaster just like shoving Linq every-where.

EDIT2: 
`public static async Task<Account> CreateAccount(string email, string password)
        {
            email = email.ToLower();
            using (var context = new AuthContext())
            {
                // Ensure account doesn't already exist with the same email
                if (context.Account.FirstOrDefault(a => a.Email == email) != null) `

It's already in an Async Task so further asyncing isn't necessary and would be generally excessive as there is finite amount of threads available.

Fabi: Little side note: async/await won't create threads in most cases. There is no guarantee that it's in a new thread. You should not care about that when using async tasks.

VoidEssy: While true you also can't guarantee it won't spawn one since when you have Task<Type> it will be allocated to a background thread, which may or may not expand threads in use. 
Ref: 
1. https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1.result?redirectedfrom=MSDN&view=netframework-4.8#System_Threading_Tasks_Task_1_Result
2. https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1.result?redirectedfrom=MSDN&view=netframework-4.8#System_Threading_Tasks_Task_1_Result
3. https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/task-based-asynchronous-programming
4. https://docs.microsoft.com/en-us/dotnet/csharp/async (ESPECIALLY!) 

> For I/O-bound code, you await an operation which returns a Task or Task<T> inside of an async method.
> 
> For CPU-bound code, you await an operation which is started on a background thread with the Task.Run method.
> 
> The await keyword is where the magic happens. It yields control to the caller of the method that performed await, and it ultimately allows a UI to be responsive or a service to be elastic.

5. https://docs.microsoft.com/en-us/dotnet/standard/threading/foreground-and-background-threads

The truth of the matter isn't that black and white i heavily recommend looking specifically into REF 4 and 5.

PS. Not trying to argue or 'prove you wrong' more like trying to point out that, it's something you can't control unless you manually create and discard threads. So it's not a magic bullet and has to be used sensibly.

Therzok: IO-bound APIs are not going to create new threads, since they use async-IO capabilities on most platforms, afaik. Async methods for DB access should be _IO bound_, so they should not be using the threadpool in this case.

I could be wrong, but I think only Mono used the threadpool for IO, .NET Framework and Core have this implemented properly.

VoidEssy: Yeah if your action is I/O Bound (GPU, Net-card etc.) No threads will be created as processors can run parallel to the CPU. But thing is the server code is mostly CPU Bound as it deals with memory and data in memory which is a CPU bound activity. In which case it uses an existing thread-pool which isn't infinite. (Not sure how much the code does calculations those are the biggest culprits in taking up background threads)

Simply put what's the difference between Async Await and a Thread.

Thread - Is built up and torn down which is very performance heavy in comparison to Async so there are few cases where now-days you should use it unless you're doing some low level infrastructure software.

Async Await - There is a background thread (As a matter of fact it's a pool) which receives the task and either return the result or gets discarded if parent process is discarded (Which threads don't do that's their flaw). Now on this one I'm not sure but if Background Thread pool starts running out the system will allocate some extra threads but those threads again don't come endlessly. Now if you have a parent process that creates 1000000s of different async tasks and it never dies (It's the core loop) and some of those tasks never complete they go into infinite loop or some dead lock occurs. You now have tasks in a Background thread not releasing resources.


I know I know this is maybe a bit overly cautious and paranoid but Murphy's law.

Also please correct me if I've any form of misconceptions in my statement.

Therzok: Using sync-blocking APIs in Task APIs is way more dangerous than using async/await multiple times. If the db query ends up having to do IO, it will release the current task's thread to the threadpool, so the changes of having threadpool starvation by queueing up multiple queries (IO is way slower than CPU, the db query has both an IO and a CPU component) is smaller, imo.

I'm going to describe an example of a similar situation:

In example, a simple `lock (lockObject)` statement will keep a thread blocked in a task that is waiting for a critical section. That means that the thread is stuck until another thread finishes working in the critical section.

In contrast, using a `SemaphoreSlim` and its `WaitAsync` will yield the current tasks' thread to the threadpool, and the execution will continue once the task is chosen to run again by the scheduler and the section is signaled.

It's a trade-off that mostly needs to be profiled actively. The threadpool scaling mechanism is well defined, and there are [means to control the maximum number of threadpool threads](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.setmaxthreads?view=netframework-4.8) that get created.

On the topic of EFCore itself, it uses ValueTask in case the data is cached - that means it usually completed synchronously - otherwise switches to an actual IO query.

> Now if you have a parent process that creates 1000000s of different async tasks and it never dies (It's the core loop) and some of those tasks never complete they go into infinite loop or some dead lock occurs. You now have tasks in a Background thread not releasing resources.

I'm not sure I follow how this state can be reached.

Therzok: Just to put this here:
https://docs.microsoft.com/en-us/ef/core/querying/async

> Asynchronous operations can also increase throughput in a web application, where the thread can be freed up to service other requests while the database operation completes

VoidEssy: Your explanation seems to make sense. And the example case I provided is something I've actually seen once occur, that is example of people using shit as magic bullets. Hence my caution of "Yes it's powerful! Yes it's awesome! but don't slap it every-where, where it's not necessary".

Such state can be quite easily achieved in Collection parsing where each iteration through data is a new task that is supposed to return but due to spaghetti code can loop endlessly within parsed data object. (Obv. this is poor execution not something that any developer should do.)

But yeah TL;DR which was obv. from get go.
"Shouldn't be an issue as long as you use your head."

Therzok: In this specific scenario, the bottleneck of not using async/await means:

* N clients request something from the server
* The server has to query the DB for the information for the N clients
* If sync API is used, all the requests will end up _blocking_ on the DB access. That means the threadpool blocks on waiting for the DB resource, so the threadpool will ramp up, increasing the number of threadpool threads.
* If async API is used, the tasks will yield the thread to the threadpool until the DB resource is ready to be accessed. Other work can be processed in the meantime.

VoidEssy: Okay yeah, that wouldn't be a problem in a regular webservice tho as whenever service contract is consumed it's a new instance and such issues don't arise. But it makes sense. I think TBH I'd spawn a Task with DB context for each client... or in general create an instance per client within the core-loop. 

But again makes sense, not arguing against it.

Issue #128, opened 2019-03-26
Character Creation Failure


badbaki:
_Aaaw Shoot..._
Character creation failure. Unsure of what exactly is causing it but here is a creation code that fails: `eNrj/cDAwMTAwMDCAAHLgZgdiEWBmBGIORkg8iAxZiBmg6qFiXFBMUhMEojdoPqKWEEqGBuYwSQLmGQFkpmLQSbM5QXpldkKMr34KEj//LO8QPL3OT4g+b+eH0wKAMlzbIJgthDIhAPCYLYIkOzgFAWzxYDks1PiYLYEkHTwAQCrURm4`

Here is the error:
```
13:14:42.8952[ERROR][Session] System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
   at NexusForever.WorldServer.Network.Message.Handler.CharacterHandler.<HandleCharacterCreate>g__GetCharacterCustomisation|2_0(Dictionary`2 customisations, UInt32 race, UInt32 sex, UInt32 primaryLabel, UInt32 primaryValue) in C:\Users\Baki\Desktop\NF-DevFork\NexusForever\Source\NexusForever.WorldServer\Network\Message\Handler\CharacterHandler.cs:line 310
   at NexusForever.WorldServer.Network.Message.Handler.CharacterHandler.HandleCharacterCreate(WorldSession session, ClientCharacterCreate characterCreate) in C:\Users\Baki\Desktop\NF-DevFork\NexusForever\Source\NexusForever.WorldServer\Network\Message\Handler\CharacterHandler.cs:line 211
   at NexusForever.Shared.Network.GameSession.HandlePacket(ClientGamePacket packet) in C:\Users\Baki\Desktop\NF-DevFork\NexusForever\Source\NexusForever.Shared\Network\GameSession.cs:line 186 System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
[...]

PR #148, opened 2019-04-26
fix for character customisation

ahoiahoi:
this will fix e.g. the missing eyes. 
this will only fix new characters


ahoiahoi:
https://github.com/NexusForever/NexusForever/issues/128

kirmmin:
I believe this has been replaced with #191. That PR resolves all customisation issues, to my knowledge. Test server has had it for a few days at this point, and no comments that customisation has been broken since.

PR #151, opened 2019-05-05
Fixing typo and also improved the runtime performance of the generation manager

tdymel:
1. Fixing the akward typo (Prio 1)
2. Processing each file is a rather independent task, so Multithreading it makes sense. This approach is rather crude and does not consider the amount of course, so it could be improved by doing some counting there. But even with this crude approach I improved the runtime by 150% on my machine.

The Extractor could also be multithreaded but it is already fast enough.


Source/NexusForever.MapGenerator/GenerationManagaer.cs
```diff
@@ -23,14 +25,20 @@ public static void Initialise()
...
 
+            List<Thread> threadList = new List<Thread>();
             foreach (WorldEntry entry in GameTableManager.World.Entries
                 .Where(e => e.AssetPath != string.Empty)
                 .GroupBy(e => e.AssetPath)
                 .Select(g => g.First())
                 .ToArray())
             {
-                ProcessWorld(entry);
+                Thread thread = new Thread(() => ProcessWorld(entry));
```
MarkWilds: Wouldn’t it be better to use TPL here? @jasoncouture 
Even if these are “long” running work?

This could potentially starve all resources and block the main thread right. It might not be a direct problem, but who knows when someone else adds more concurrent or parallel work it will be?

Therzok: Using async IO APIs here combined with Task.WhenAll would be way better. Agreed with @MarkWilds

MarkWilds: Indeed, not to mention ProcessWorld is IO bound right? (havent looked inside the code)
Meaning operating system IO optimizations (interrupts, etc...) can be used by utilizing IO async methods.

tdymel: I changed it to use TPL instead.

PR #152, opened 2019-05-05
Player starting locations

ahoiahoi:


Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
```diff
@@ -282,18 +282,42 @@ public static void HandleCharacterCreate(WorldSession session, ClientCharacterCr
...
+                {
+                    // veteran
+                    if (((Race)creationEntry.RaceId == Race.Human && (Faction)creationEntry.FactionId == Faction.Exile) || (Race)creationEntry.RaceId == Race.Granok)
+                        worldLocation = GameTableManager.WorldLocation2.GetEntry(1594);
+                    else if ((Race)creationEntry.RaceId == Race.Chua || (Race)creationEntry.RaceId == Race.Draken)
+                        worldLocation = GameTableManager.WorldLocation2.GetEntry(8223);
+                    else if (((Race)creationEntry.RaceId == Race.Human && (Faction)creationEntry.FactionId == Faction.Dominion) || (Race)creationEntry.RaceId == Race.Mechari)
+                        worldLocation = GameTableManager.WorldLocation2.GetEntry(18987);
+                    else // Aurin / Modresh
+                        worldLocation = GameTableManager.WorldLocation2.GetEntry(37015);
```
Rawaho: Replacing this entire condition block with pattern matching and a switch statement would be much cleaner.

Source/NexusForever.WorldServer/Game/AssetManager.cs
```diff
@@ -106,6 +108,19 @@ private static void CacheItemDisplaySourceEntries()
...
+        private static void CacheXpPerLevel()
+        {
+            var entries = new Dictionary<uint, byte>();
+            foreach (XpPerLevelEntry entry in GameTableManager.XpPerLevel.Entries)
+            {
+                if (!entries.ContainsKey(entry.MinXpForLevel))
+                    entries.Add(entry.MinXpForLevel, (byte)entry.Id);
+            }
+
+            xpPerLevel = entries.ToImmutableDictionary(e => e.Key, e => e.Value);
```
Rawaho: I realise the rest of the asset manager doesn't do this (it should) but using an immutable dictionary builder would be more efficient here.

Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
ahoiahoi: this part is now out sourced to the DB

Source/NexusForever.WorldServer/Game/AssetManager.cs
ahoiahoi: done

Source/NexusForever.WorldServer/Game/AssetManager.cs
```diff
@@ -61,7 +63,7 @@ private static void CacheCharacterCustomisations()
 
         private static void CacheInventoryEquipSlots()
         {
-            var entries = new Dictionary<ItemSlot, List<EquippedItem>>();
+            Dictionary<ItemSlot, List<EquippedItem>> entries = new Dictionary<ItemSlot, List<EquippedItem>>();
```
Therzok, line 75: This should also be ported to ImmutableDictionary.Builder as methods below?

Source/NexusForever.WorldServer/Game/AssetManager.cs
```diff
@@ -106,6 +108,19 @@ private static void CacheItemDisplaySourceEntries()
             itemDisplaySourcesEntry = entries.ToImmutableDictionary(e => e.Key, e => e.Value.ToImmutableList());
         }
 
+
+        private static void CacheXpPerLevel()
+        {
+            ImmutableDictionary<uint, byte>.Builder entries = ImmutableDictionary.CreateBuilder<uint, byte>();
+            foreach (XpPerLevelEntry entry in GameTableManager.XpPerLevel.Entries)
+            {
+                if (!entries.ContainsKey(entry.MinXpForLevel))
```
Therzok: avoid double lookup by doing: `entries[entry.MinXpForLevel] = (byte)entry.Id`?

Source/NexusForever.WorldServer/Game/AssetManager.cs
```diff
@@ -130,5 +145,12 @@ public static ImmutableList<ItemDisplaySourceEntryEntry> GetItemDisplaySource(ui
         {
             return itemDisplaySourcesEntry.TryGetValue(itemSource, out ImmutableList<ItemDisplaySourceEntryEntry> entries) ? entries : null;
         }
+
+        public static byte Xp2Level(uint xp)
+        {
+            return xpPerLevel
+                .OrderByDescending(x => x.Key)
```
Therzok: Instead of always reordering by descending, wouldn't it make more sense to do it just once at cache time?

Source/NexusForever.WorldServer/Game/AssetManager.cs
Fabi: Only if you can be sure that it's not overwritten.

Source/NexusForever.WorldServer/Game/AssetManager.cs
ahoiahoi: not as easily because of the lists in it. it would just stretch the code unnecessarily in my opinion

Source/NexusForever.WorldServer/Game/AssetManager.cs
ahoiahoi: yes I did intend that, but did not find a method to store an ordered dictionary

Source/NexusForever.WorldServer/Game/AssetManager.cs
Therzok: Why does it need to be a dictionary? It could simply be a `List<(uint, byte)>` that's constructed with a comparer that compares the XP amount. Then, it could be a binary search for where the value should go in.

Source/NexusForever.WorldServer/Game/AssetManager.cs
Therzok: Why does it need to be a dictionary? It could simply be a `List<(uint, byte)>` that's constructed with a comparer that compares the XP amount. Then, it could be a binary search for where the value should go in.

Source/NexusForever.WorldServer/Game/AssetManager.cs
Therzok: I'm not sure why it needs to be a dictionary though, see https://github.com/NexusForever/NexusForever/pull/152/files#r285359371

Source/NexusForever.WorldServer/Game/AssetManager.cs
ahoiahoi: I thought this would give good performance over a list... am I wrong?
I expect this function to be called very frequently

Therzok: No, imo, in this case, a List<T> should be used, with log2n insertion via 2complement on a negative result 

https://docs.microsoft.com/en-us/dotnet/api/system.array.binarysearch?view=netframework-4.8

Fabi: You are overcomplicating things with that

Therzok: Tl;dr only contains (exact match) is reason to use Dictionary. In this case, we're still doing O(n) iteration to find minimum index.

Tbh, easiest option would he to use BinarySearch

jasoncouture: I agree with the twos complement of binary search, but only when the object is used a lot.
Ease of understanding the code is better when performance is not required.

ahoiahoi: implemented the binary search - hope this really does perform better than the dictionary

Database/Updates/Character/2019_05_06_realmconfig.sql
```diff
@@ -0,0 +1,90 @@
+CREATE TABLE IF NOT EXISTS `realmconfig` (
+    `id` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0',
```
kirmmin, line 2: Is this increased by a command in game, or manually adding entries? If manually adding entries, it might make sense to change this to an AUTO-INCREMENT default?

kirmmin: Additionally, how does this realm config compare to the JSON configuration we're using for other values? I understand that this offers more flexibility, but it might be better to have the "active" realm config be contained in the JSON file and forego having an extra table?

Database/Updates/Character/2019_05_06_realmconfig.sql
```diff
@@ -0,0 +1,90 @@
...
+    `active` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0',
+    PRIMARY KEY (`id`)
+);
+
+CREATE TABLE IF NOT EXISTS `realmconfig_starting_location` (
```
kirmmin, line 7: Is it worthwhile adding level into these entries? I can see it was hardcoded in the CharacterHandler but as you've gone to the trouble of calculating Xp to Level, it might make sense to update the starting levels per starting entry?

ahoiahoi: not sure what you are refering to as above/below is not level/xp related

Database/Updates/Character/2019_05_06_realmconfig.sql
ahoiahoi: could be auto inc, yes - we don't currently use auto inc at all, though
the active choosen config could be configured in the json - I think it will be easier for some kind of server admin frontend to have the values in the database

kirmmin: I understand; My concern is that @Rawaho suggested to put any configuration into the JSON file, which I've been doing (where appropriate), and didn't want to put a bunch of configuration in different places.
For multiple entries (like the actual custom locations), that should be in the DB. But if you're selecting an Entry Set, I think it might make more sense to have that in the JSON.

Just my opinion, I was just clarifying my point before. Might be something Rawaho has a different opinion on, and yours is the best way to go :)

Database/Updates/Character/2019_05_06_realmconfig.sql
kirmmin: Sorry, I put this in the wrong line, probably. I was referring to this hardcoded value: https://github.com/NexusForever/NexusForever/pull/152/files#diff-2b0a4f087ba8ac9b03a47b11ccd20ddcR237

Might make sense to have that associated with the starting location, or perhaps another table like "realmconfig_starting_data" that has things like level or gear, or whatever, when it's custom?

ahoiahoi: you mean the "creationLevel = 1"? that is just initialised to 1, it is effectivly assigned later based on the xp from the CharacterCreation tbl
or do I missunderstand something?

Database/Updates/Character/2019_05_06_realmconfig.sql
ahoiahoi: In my opinion it fits either way - I'll leave it up to Rawaho to deceide upon merge or change request

Database/Updates/Character/2019_05_06_realmconfig.sql
kirmmin: Oh, I see what you mean. I misunderstood what was going on. You can ignore this level-related comment :)

PR #153, opened 2019-05-06
Automatic setup for mac/linux or linux subsystem users

tdymel:
These scripts are supposed to help newcomers into the project. 
The setup script sets up the server in such a way, that no build files can be accidentally added when committing changes.

Huge parts of the setup script are stolen from Ahoi's wiki entries. But it was a little cryptic, so I had to squeeze out some information of Ahoi, such that things can get setup properly here.

This script should simplify the entry for most users into the project.


Tools/Linux/run.sh
```diff
@@ -0,0 +1,12 @@
+# Making sure that paths are correct
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )";
+cd $DIR/Source;
```
ahoiahoi: you should 
a) always use ${} as variable accessor
b) not use semicolon at the end
c) use something like 'cd ${DIR} || exit "something wrong"'

Tools/Linux/setup.sh
```diff
@@ -0,0 +1,56 @@
...
+DNS="user=$DB_USER";
+
+if [ "$DB_PW" != "" ]; then
+        DNS="user=$DB_USER;password=$DB_PW";
+fi;
+
+# Making sure that paths are correct
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/../../";
+cd $DIR;
+
```
ahoiahoi: see above
also, the script may not be invoked from the directory it resides in

Tools/Linux/setup.sh
```diff
@@ -0,0 +1,56 @@
...
+# Cleaning
+rm -rf ${DIR}/Source/NexusForever.AuthServer/bin;
+rm -rf ${DIR}/Source/NexusForever.StsServer/bin;
+rm -rf ${DIR}/Source/NexusForever.WorldServer/bin;
+
```
ahoiahoi: never use "rm -rf" in a script without making sure the the variable is not empty

Tools/Linux/kill.sh
```diff
@@ -0,0 +1,2 @@
+pkill -f NexusForever.StsServer
+pkill -f NexusForever.AuthServer
```
ahoiahoi: add linefeed, and what about the world server?

tdymel: The worldserver is not started as decoppled to the prompt, i. e. with &

Tools/Linux/setup.sh
tdymel: At the end u can see a "/../../" appended to the directory. This makes sure that the source directory is correct. I added an empty check tho

tdymel:
Usage:
Setup:
./nexus.sh -s game_path table_path db_user [db_pw]
Kill:
./nexus.sh -k
Run:
./nexus.sh -r

PR #155, opened 2019-05-23
Feature: Role Based Access Control - WIP

kirmmin:
**Features**
- Minimum Status for all commands and subcommands
- Configurable via WorldServer.json
- GM Flag support in chat (not modifiable at this time)


kirmmin:
Migrated this PR to use an RBAC model. This adds the following capabilities:
- Accounts can now have roles tied to them. There's some starter ones:
   - Player
   - Sandbox Player
   - Game Master
   - Administrator
- Accounts can have specific permissions tied directly to them, regardless of any roles associated to it.
- Roles can be created (in the DB at the moment)
- Permissions can be added to Roles (in the DB at the moment)
   - Permissions are hard-coded in the server as part of an Enum. (@Hallowisp, could be a good idea to add this to the Wiki as API documentation if this PR gets accepted)

PR #158, opened 2019-05-24
Bug Fix: Teleporting & Ability use while mounted

### LINKED ISSUE #117: Issues while mounted
- Spell effect animations play while mounted.
- No way of tracking user mounted state, currently.
   - Do we plan on adding player flags to the class? I know some will get recognised in game by LUA by the entity state, but it's good to track this server side, too.

Extends #113

### LINKED ISSUE #118: Additional issues while mounted
Bike mounts like the Uniblades or lever mounts like Igneous Strider have the incorrect player appearance pose. 

Teleporting while mounted: Depending on zone the player will either end up with the player appearance bug in #113 or end up teleported to XYZ coordinates of 0, 0, 0 for given target zoneID.

To reproduce:
1- Mount up
2- !go Thayd or !go Illium
3- Go skydiving for a bit

Extends #117 and #113

kirmmin:
Fixes #117 
Fixes #118

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Entity/Player.cs
+++ b/Source/NexusForever.WorldServer/Game/Entity/Player.cs
@@ -490,6 +490,8 @@ namespace NexusForever.WorldServer.Game.Entity
                 pet?.RemoveFromMap();
                 VanityPetGuid = null;
             }
+            
+            DestroyDependents();
 
             base.OnRemoveFromMap();
 
@@ -666,6 +668,9 @@ namespace NexusForever.WorldServer.Game.Entity
                 VanityPet pet = GetVisible<VanityPet>(VanityPetGuid.Value);
                 vanityPetId = pet?.Creature.Id;
             }
+            
+            if (VehicleGuid != 0u)
+                Dismount();
 
             var info = new MapInfo(entry, instanceId, residenceId);
             pendingTeleport = new PendingTeleport(info, vector, vanityPetId);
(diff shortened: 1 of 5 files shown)


Source/NexusForever.WorldServer/Game/Entity/UnitEntity.cs
```diff
@@ -73,6 +73,17 @@ public void CastSpell(SpellParameters parameters)
             if (parameters == null)
                 throw new ArgumentNullException();
 
+            if (GetType() == typeof(Player))
```
ahoiahoi: might want to use "if (this is Player player)" instead

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -536,6 +538,42 @@ public void TeleportTo(WorldEntry entry, Vector3 vector, uint instanceId = 0u, u
...
+        }
+
+        /// <summary>
+        /// Dismounts this <see cref="Player"/> from a vehicle that it's attached to
+        /// </summary>
+        public void Dismount()
+        {
+            if (VehicleGuid != 0u)
+            {
+                Vehicle vehicle = GetVisible<Vehicle>(VehicleGuid);
```
Fabi, line 802: Possible null reference?

kirmmin:
Rebased today to the latest `develop` branch.

Went through the packets today and discovered that there was an inconsistency with retail sniffs when dismounting, specifically around the fact that the EntityCommand packet update was not sent on Retail. The sniffs I have with a player being removed from a mount never had a `ServerEntityCommand` packet sent after removal.
Removing that prevented the client "responding" to the EntityCommand instruction with it's own request.

This is the line I removed: https://github.com/NexusForever/NexusForever/pull/158/files#diff-c73a1edb9532580b34666eac934fb1daL234

I tested Dismounting, Dismount from Teleport, Logging out, and ALT+F4-ing client. None of these events triggered the exception we were seeing before (or other exceptions for that matter). I tested this with 2 clients open, as well, and was able to see the appropriate updates to both clients, so I assume the Broadcast was just an oversight. This may well be finally resolved :)

Removing WIP

PR #161, opened 2019-05-27
Feature: In Game Day/Night Cycle

kirmmin:


Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -536,6 +538,20 @@ public void TeleportTo(WorldEntry entry, Vector3 vector, uint instanceId = 0u, u
             RemoveFromMap();
         }
 
+        /// <summary>
+        /// Used to send the current in game time to this player
+        /// </summary>
+        private void SendInGameTime()
+        {
+            uint lengthOfInGameDayInSeconds = ConfigurationManager<WorldServerConfiguration>.Config.LengthOfInGameDay <= 0 ? 12600 : ConfigurationManager<WorldServerConfiguration>.Config.LengthOfInGameDay; // Live servers were 3.5h per in game day
+            double timeOfDay = DateTime.Now.Subtract(DateTime.Today).TotalSeconds / lengthOfInGameDayInSeconds % 1;
```
Therzok: Because of the modulo and the fact that 12600 is not directly a divisor of 86400 (seconds in a day) the time of day will leap at the end of a day.

| Clock time | TimeSpan.TotalSeconds | secondsInDay     |
|------------|-----------------------|------------------|
| 23:59:59   | 86399                 | 0.85706349206349 |
| 00:00:00   | 0                     | 0                |

We should use a fixed time to get the seconds passed from. Either unix epoch or time since server start.

`DateTime.Now.Substract(DateTime.Today)` has 2 nasty bugs waiting to happen:

* `DateTime.Now` will return a different value after a DST change. So time will skip again going for one hour. Use `DateTime.UtcNow`.
* There is a possibility the 2 calls (`DateTime.Now` and `DateTime.Today`) are called on different seconds, giving us the first described scenario, but with a negative time duration:

```
DateTime.Now.Substract(DateTime.Today);
^ 23:59:59             ^ 00:00:00
```

Proposal:
```csharp
double timeOfDay = DateTime.UtcNow.Subtract(DateTime.UnixEpoch).TotalSeconds / lengthOfInGameDayInSeconds % 1;
```

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -536,6 +538,20 @@ public void TeleportTo(WorldEntry entry, Vector3 vector, uint instanceId = 0u, u
...
+            Session.EnqueueMessageEncrypted(new ServerTimeOfDay
+            {
+                TimeOfDay = (uint)(timeOfDay * TimeSpan.FromDays(1).TotalSeconds),
```
Therzok, line 554:
```csharp
const long SecondsInADay = TimeSpan.TicksPerDay / TimeSpan.TicksPerSecond;
...
TimeOfDay = (uint)(timeOfDay * SecondsInADay);
```

Source/NexusForever.WorldServer/Game/Entity/Player.cs
kirmmin: Really good point. I've mentioned the `UtcNow` issue on the guild branch, and this runs the risk of being out of sync with the rest of the server, so I'll continue to use `DateTime.Now` until we read UTC dates from the DB - unless @Rawaho tells me we already do :)

I'll add in the `UnixEpoch`, that's a really good idea. Thanks for the math.

Source/NexusForever.WorldServer/Game/Entity/Player.cs
kirmmin: This code was Rawaho's suggestion - what's the benefit of using your suggestion?

Therzok: And on that note, maybe we could have another constant too, for the default game day duration:
```csharp
// 12600 - Live servers were 3.5h per in game day
const double DefaultSecondsInAGameDay = 3.5 * TimeSpan.TicksPerHour / TimeSpan.TicksPerSecond; 
```

Therzok: Well, we'd not do useless math on the CPU of converting 1 day to ticks and then get the seconds. :D

kirmmin: I wanted to avoid adding more variables, but I've make the assignment clearer now, and used an hour expression so it's a bit easier to read.

PR #163, opened 2019-05-28
fix for the `item` table changes by the mail update

ahoiahoi:
fix for the FK


Rawaho:
What exactly is the issue here?
The FK is still required and having a nullable FK is valid.

ahoiahoi:
the problem is that the FK already exists
https://github.com/NexusForever/NexusForever/blob/master/Database/Updates/Character/2018_10_18_item.sql

PR #165, opened 2019-05-30
Gameplayer/Features: Transmat Support & Client Interaction - WIP

kirmmin:
Contributes towards #306 

**Features**
- Binds players to any Transmat location
- Teaches the necessary Recall spell on bind, if the player does not know it
- Bind locations are saved to the database per player
- Client Interaction Events have basic support, allowing for things like the Minigames where you match colours, or the "tap quickly/slowly or at specific times", to be supported.
   - Validation of these events is currently not implemented. We rely on the Client as the source of truth which is not ideal.
- Journal / Key cast times have been implemented. In fact, any Simple entity that has an interaction event that requires a cast time of some sort should fire the spell cast now.
   - Note: This does not mean every interaction "works", just that it will cast the associated spell.

**Findings**
- "Activate" spell effect is probably how Carbine tied spells to scripts on the server. Whenever I see Activate, there's never any data associated with it in the table, suggesting that the server stored this piece.
- "ActorPropId" entity property forces that entity to assume the map prop's location & model, and toggles the model's state/effects when the entity is told to do something.
   - For a proof of concept, go to Tremor Ridge, Algoroc and check the Transmat there. This is setup exactly as Live was (based on the sniffed data I have) and you can see, between the `BindPoint.cs` and in-game visuals, how things work.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/AssetManager.cs
+++ b/Source/NexusForever.WorldServer/Game/AssetManager.cs
@@ -1,11 +1,14 @@
 ﻿using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
+using System.Numerics;
 using System.Reflection;
 using NexusForever.Database.World.Model;
 using NexusForever.Shared;
 using NexusForever.Shared.Database;
 using NexusForever.Shared.GameTable;
 using NexusForever.Shared.GameTable.Model;
+using NexusForever.WorldServer.Game.Entity;
 using NexusForever.WorldServer.Game.Entity.Static;
 using NexusForever.WorldServer.Game.Quest.Static;
 
@@ -38,6 +41,7 @@ namespace NexusForever.WorldServer.Game
 
         private ImmutableDictionary<ItemSlot, ImmutableList<EquippedItem>> equippedItems;
         private ImmutableDictionary<uint, ImmutableList<ItemDisplaySourceEntryEntry>> itemDisplaySourcesEntry;
+        private ImmutableDictionary<ushort, Location> bindPointLocations;
 
         private ImmutableDictionary</*zoneId*/uint, /*tutorialId*/uint> zoneTutorials;
         private ImmutableDictionary</*creatureId*/uint, /*targetGroupIds*/ImmutableList<uint>> creatureAssociatedTargetGroups;
@@ -58,6 +62,7 @@ namespace NexusForever.WorldServer.Game
             CacheItemDisplaySourceEntries();
             CacheTutorials();
             CacheCreatureTargetGroups();
+            CacheBindPointPositions();
         }
 
         private void CacheCharacterCustomisations()
@@ -77,7 +82,7 @@ namespace NexusForever.WorldServer.Game
 
         private void CacheInventoryEquipSlots()
         {
-            var entries = new Dictionary<ItemSlot, List<EquippedItem>>();
+            var entries = ImmutableDictionary.CreateBuilder<ItemSlot, List<EquippedItem>>();
             foreach (FieldInfo field in typeof(ItemSlot).GetFields())
             {
                 foreach (EquippedItemAttribute attribute in field.GetCustomAttributes<EquippedItemAttribute>())
@@ -95,7 +100,7 @@ namespace NexusForever.WorldServer.Game
 
         public void CacheInventoryBagCapacities()
         {
-            var entries = new Dictionary<InventoryLocation, uint>();
+            var entries = ImmutableDictionary.CreateBuilder<InventoryLocation, uint>();
             foreach (FieldInfo field in typeof(InventoryLocation).GetFields())
             {
                 foreach (InventoryLocationAttribute attribute in field.GetCustomAttributes<InventoryLocationAttribute>())
@@ -105,7 +110,7 @@ namespace NexusForever.WorldServer.Game
                 }
             }
 
-            InventoryLocationCapacities = entries.ToImmutableDictionary();
+            InventoryLocationCapacities = entries.ToImmutable();
         }
 
         private void CacheItemDisplaySourceEntries()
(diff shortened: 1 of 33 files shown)


kirmmin:
Updated the BindPoint positions to be stored in the asset manager and will only store them if the Entity exists in the DB.

Additionally, I tidied up the way in which the casting process works. It felt a bit clunky before, creating a ClientSideInteraction class which then cast the spell, etc. This way, the Entity casts the spell, which actually makes it more accurate because, while testing, it appears that the `CasterId` MUST be the entity the player interacted with.

Source/NexusForever.WorldServer/Game/AssetManager.cs
```diff
@@ -110,6 +116,33 @@ private static void CacheItemDisplaySourceEntries()
             itemDisplaySourcesEntry = entries.ToImmutableDictionary(e => e.Key, e => e.Value.ToImmutableList());
         }
 
+        private static void CacheBindPointPositions()
+        {
+            var entries = new Dictionary<ushort, Location>();
```
Therzok:
```suggestion
            var entries = ImmutableDictionary.CreateBuilder<ushort, Location>();
```

Source/NexusForever.WorldServer/Game/AssetManager.cs
```diff
@@ -110,6 +116,33 @@ private static void CacheItemDisplaySourceEntries()
...
+                if (worldEntry == null)
+                    continue;
+
+                Location bindPointLocation = new Location(worldEntry, new Vector3(entityEntry.X, entityEntry.Y, entityEntry.Z), new Vector3(entityEntry.Rx, entityEntry.Ry, entityEntry.Rz));
+
+                if (!entries.ContainsKey(entryId))
+                    entries.Add(entryId, bindPointLocation);
+            }
+
+            bindPointLocations = entries.ToImmutableDictionary(e => e.Key, e => e.Value);
```
Therzok:
```suggestion
            bindPointLocations = entries.ToImmutable();
```

Source/NexusForever.WorldServer/Game/AssetManager.cs
kirmmin: What's the benefits of doing it this way? I copied the other methods in the `AssetManager`, so if I'm going to break the mold, I should probably update the others, and will need to know why I would do it this way :)

Rawaho: This is more efficient, the builders are already used in multiple places, anything not using them needs to be updated.

kirmmin: Rawaho has spoken, so I have some updates to do ;) Thanks for the pointer, Therzok.

Therzok: @kirmmin Basically, ImmutableDictionary (and most other ImmutableCollections other than Array), use a Root+Nodes implementation, technically being trees. That allows for reusing of nodes across multiple instances, referencing to each version as a specific root + node chain.

Modifying an immutable collection via it's APIs will cause mutation (clone modified subset) and then freezing. An ImmutableDictionary.Builder internally does all the mutations needed, and only applies one freeze at the end.

Using a normal `Dictionary<TKey, TValue>`, it is implemented internally as a bucket array with TKey and TValue as respective bucket pairs.

So, when using an ImmutableDictionary builder, you're basically constructing the tree nodes from the get go, avoiding creating an extra Dictionary instance and constructing that.

Using Dictionary -> ToImmutable(), we're processing twice:
`new Dictionary() -> Add N KeyValuePairs -> new ImmutableDictionary() -> Mutate once with all pairs from dictionary -> snapshot`

The mutation itself from above is basically what an ImmutableDictionary builder would do. So we have higher computation costs.

Using ImmutableDictionary.Builder:
`new Builder -> Add N KeyValuePairs -> finalize internal structure to ImmutableDictionary with data`

kirmmin: I've made some changes following your suggestions. I've updated the other Caching methods to use the same methodology, but I'd like somebody else's eyes on that as I'm unfamiliar with this :)

Source/NexusForever.WorldServer/Game/AssetManager.cs
```diff
@@ -46,11 +51,12 @@ public static void Initialise()
             CacheInventoryEquipSlots();
             CacheInventoryBagCapacities();
             CacheItemDisplaySourceEntries();
+            CacheBindPointPositions();
         }
 
         private static void CacheCharacterCustomisations()
         {
-            var entries = new Dictionary<uint, List<CharacterCustomizationEntry>>();
+            var entries = ImmutableDictionary.CreateBuilder<uint, List<CharacterCustomizationEntry>>();
```
Therzok: On this case, I would go with:

```suggestion
            var entries = ImmutableDictionary.CreateBuilder<uint, ImmutableList.Builder<CharacterCustomizationEntry>>();
```

Therzok: And then, characterCustomisations can just be set like:
```csharp
characterCustomisations = entries.ToImmutable ();
```

Source/NexusForever.WorldServer/Game/AssetManager.cs
```diff
@@ -93,12 +99,12 @@ public static void CacheInventoryBagCapacities()
...
             }
 
-            InventoryLocationCapacities = entries.ToImmutableDictionary();
+            InventoryLocationCapacities = entries.ToImmutable();
         }
 
         private static void CacheItemDisplaySourceEntries()
         {
-            var entries = new Dictionary<uint, List<ItemDisplaySourceEntryEntry>>();
+            var entries = ImmutableDictionary.CreateBuilder<uint, List<ItemDisplaySourceEntryEntry>>();
```
Therzok: Same here, with converting to ImmutableList.

kirmmin:
Added WIP to this PR because I'd like #213 merged first and then I'll migrate this code to that set of code.

kirmmin:
Note for self: This needs to be rewritten. ClientSideInteraction is a cast method and not based on entity interaction (some entity interactions are only "Normal" casts, like the Quest activate spell). Will fix this as part of the after the Spell Mega Update.

kirmmin:
CSI Updates are part of https://github.com/NexusForever/NexusForever/tree/spells-v2-gr

Once that is merged, I will come back and refactor this branch to only deal with the Bindpoint functionality.

PR #166, opened 2019-05-31
Feature: Basic tutorial trigger for zone change

kirmmin:


kirmmin:
Refactored to have zone tutorial IDs and associated triggers stored in the DB and using `AssetManager` as the Data Provider.

PR #167, opened 2019-05-31
Fix: Item location duplication fix

kirmmin:


kirmmin:
Updated to use `PendingCreate` methodology as Rawaho requested.

PR #169, opened 2019-06-01
Feature: World Prop usage by entities, Doors proof of concept

kirmmin:


Source/NexusForever.WorldServer/Command/Handler/DoorCommandHandler.cs
```diff
@@ -0,0 +1,74 @@
...
+            uint searchRange = 10;
+            if (parameters.Length > 0)
+                searchRange = uint.Parse(parameters[0]);
+
+            context.Session.Player.Map.Search(
+                context.Session.Player.Position,
+                searchRange,
+                new SearchCheckRange(context.Session.Player.Position, searchRange, context.Session.Player),
+                out List<GridEntity> intersectedEntities
+            );
```
Rawaho: There is no need to do a grid search for nearby doors, using visible entities will be enough since there should never be a need to open doors that the player can't see.

Source/NexusForever.WorldServer/Command/Handler/DoorCommandHandler.cs
```diff
@@ -0,0 +1,74 @@
...
```
Rawaho: Same as above.

Source/NexusForever.WorldServer/Command/Handler/DoorCommandHandler.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            foreach(var entity in intersectedEntities)
+            {
+                if (entity is Door door)
+                {
+                    context.SendMessageAsync($"Trying to open door {door.Guid}");
+                    door.OpenDoor(context.Session.Player);
+                }
+            }
```
Rawaho: This isn't really a valid comment any more since you should remove the grid search but create a new search check if the future that only selects door entities rather then doing it after the fact.

Source/NexusForever.WorldServer/Game/Entity/Door.cs
```diff
@@ -0,0 +1,84 @@
...
+        public void OpenDoor(Player player, bool openForAll = true)
+        {
+            var doorEmote = new ServerEmote
+            {
+                Guid = Guid,
+                StandState = 4
+            };
+
+            SetStat(Stat.StandState, 4);
+            if (openForAll)
```
Rawaho: This will cause a desync of information between clients and the server if `openForAll` is false.
You are changing the state but only sending the result to a single player, this single player will see the door open, any new players coming into range will see it open but any characters that were nearby but haven't moved will see it closed.

Source/NexusForever.WorldServer/Game/Entity/Door.cs
```diff
@@ -0,0 +1,84 @@
...
+        public void CloseDoor(Player player, bool openForAll = true)
+        {
+            var doorEmote = new ServerEmote
+            {
+                Guid = Guid,
+                StandState = 3
+            };
+
+            SetStat(Stat.StandState, 3);
+            if (openForAll)
```
Rawaho: Same as above but in reverse.

Rawaho (changes requested):

Source/NexusForever.WorldServer/Command/Handler/DoorCommandHandler.cs
kirmmin: I've created a new Search Check, and I've also added a `GetVisibleEntities<T>` method to return all visible entities of a type.

PR #170, opened 2019-06-02
Commands: Currency handler

kirmmin:


Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+            context.SendMessageAsync($"Granted {amount} to currencyId {currencyId}.");
+            return Task.CompletedTask;
+        }
+
+        [SubCommandHandler("list", "Lists currency IDs and names")]
+        public Task ListSubCommand(CommandContext context, string command, string[] parameters)
+        {
+            List<string> Currencies = new List<string>
+            {
+                "Credits (aka Money!) - currencyId (1)",
```
Therzok: Might be easier to just use game table data for this?

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+        }
+
+        [SubCommandHandler("add", "currencyId amount - Adds currency to character.")]
+        public Task AddSubCommand(CommandContext context, string command, string[] parameters)
+        {
+            if (parameters.Length != 2)
+                return Task.CompletedTask;
+
+            var currencyId = byte.Parse(parameters[0]);
```
Therzok: `Enum.Parse (typeof(CurrencyType), parameters[0])`

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
MaxtorCoder: If there even is a GameTable for it.

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
MaxtorCoder: This works just as well. No need to make it more complicated or add extra code while it achieves the same thing.

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
Therzok: There is [one](https://github.com/NexusForever/NexusForever/blob/037455141b689a3644cc2d0d5bc35702de1fda40/Source/NexusForever.Shared/GameTable/GameTableManager.cs#L170)!

MaxtorCoder: It is not a lot of code tho, so is it really necassary?

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
Therzok: Well, there is no check for currency < 0, so we would end up with an argumentnullexception [here](https://github.com/NexusForever/NexusForever/blob/fe935d79769091aa2356984df11d816cfdc40d53/Source/NexusForever.WorldServer/Game/Entity/CurrencyManager.cs#L85).

Was thinking type safety would help us, but it returns an object here, not validating, so yeah.

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+            var amount = uint.Parse(parameters[1]);
+
+            if (currencyId > 10 || currencyId == 8 || amount > 100000000)
```
Therzok:
```suggestion
            if (currencyId < 0 || currencyId > 10 || currencyId == 8 || amount > 100000000)
```

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
Therzok: It simplifies the code and removes hardcoded values. The request in the other suggestion above was to simplify code, which is the intent of this suggestion.

```csharp
foreach (CurrencyTypeEntry currency in GameTableManager.CurrencyType)
    context.SendMessageAsync($"{currency.Id} - {currency.Description}");
```

MaxtorCoder: Hmm. Good one. I approve.

kirmmin:
Thanks for all the feedback. I've updated the handler following the reviews, and also fixed a couple issues in the Currency Manager.

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+        [SubCommandHandler("add", "currencyId amount - Adds currency to character.")]
+        public Task AddSubCommand(CommandContext context, string command, string[] parameters)
+        {
+            if (parameters.Length != 2)
+            {
+                context.SendMessageAsync("Parameters are invalid. Please try again.");
+                return Task.CompletedTask;
+            }
+
+            if(!Enum.TryParse(parameters[0], out CurrencyType currencyId))
```
Therzok: This does not return false for stuff like -1.

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            context.Session.Player.CurrencyManager.CurrencyAddAmount(currencyId, amount, true);
+            return Task.CompletedTask;
+        }
+
+        [SubCommandHandler("list", "Lists currency IDs and names")]
+        public Task ListSubCommand(CommandContext context, string command, string[] parameters)
+        {
+            foreach (int currencyType in (int[])Enum.GetValues(typeof(CurrencyType)))
```
Therzok: Again, why not use gametable? It already has the description text we need.

kirmmin: Mainly due to the fact that we use the Enum everywhere else. The only place that uses the GameTable is the manager itself

MaxtorCoder: True, but it would be better to do it via the GameTable.

kirmmin: Would you then suggest using the GameTable everywhere? A gametable lookup everywhere will be far slower than an Enum.

kirmmin: I've just realised this is on the list command :P hah. 
I didn't want to do a lookup before we go to the CurrencyManager because there might be custom rules in the future with disabled currencies that will likely be managed by the CurrencyManager.

For the List command, I'm okay with using the GameTable. It's early here, so take it easy ;)

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+        public Task AddSubCommand(CommandContext context, string command, string[] parameters)
+        {
+            if (parameters.Length != 2)
+            {
+                context.SendMessageAsync("Parameters are invalid. Please try again.");
+                return Task.CompletedTask;
+            }
+
+            bool currencyParsed = sbyte.TryParse(parameters[0], out sbyte currencyId);
+            if (!currencyParsed || currencyId <= 0 || !Enum.IsDefined(typeof(CurrencyType), (CurrencyType)currencyId))
```
Therzok: Minor nit:
```suggestion
            if (!currencyParsed || !Enum.IsDefined(typeof(CurrencyType), (CurrencyType)currencyId))
```

Therzok (approved):

kirmmin:
Switched to GameTable after speaking with @Rawaho. Reflection is much slower and that's what I was using, sorry for all the confusion :) Hopefully, now, things'll be better :)

Source/NexusForever.WorldServer/Command/Handler/CurrencyCommandHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            CurrencyTypeEntry currencyEntry = GameTableManager.CurrencyType.GetEntry(currencyId);
+            if (currencyEntry == null)
+            {
+                context.SendMessageAsync("Invalid currencyId. Please try again.");
+                return Task.CompletedTask;
+            }
+
+            if(!uint.TryParse(parameters[1], out uint amount))
```
Therzok: Styling:

```suggestion
            if (!uint.TryParse(parameters[1], out uint amount))
```

PR #171, opened 2019-06-03
Fix: Currency Change packet naming

kirmmin:


Source/NexusForever.WorldServer/Network/Message/Model/ServerCombatReward.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Used by <see cref="CurrencyManager"/>, and also used during Combat to provide Momentum Boosts.
+    /// </summary>
+    [Message(GameMessageOpcode.ServerCombatReward)]
+    public class ServerCombatReward : IWritable
+    {
+        public byte Stat { get; set; }
+        public ulong NewValue { get; set; }
+        public uint CombatRewardId { get; set; } = 0;
+        public uint TargetUnitId { get; set; } = 0;
```
Rawaho: There is no need to specify the value for a property if it's already the default value.

PR #174, opened 2019-06-04
Feature: Stance/Innate set & save

kirmmin:
Also resolves some endless loop spell casting on bad innate abilities


Source/NexusForever.WorldServer/Network/Message/Model/ClientInnateChange.cs
```diff
@@ -0,0 +1,10 @@
+using NexusForever.Shared.Network;
+using NexusForever.Shared.Network.Message;
+
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    [Message(GameMessageOpcode.ClientInnateChange)]
+    public class ClientInnateChange : ServerPlayerInnate
```
Rawaho: A client packet should never inherit a server packet and vice versa.

Source/NexusForever.WorldServer/Network/Message/Model/ServerPlayerInnate.cs
```diff
@@ -0,0 +1,22 @@
...
+using NexusForever.Shared.Network.Message;
+
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    [Message(GameMessageOpcode.ServerPlayerInnateSet)]
+    public class ServerPlayerInnate : IWritable, IReadable
+    {
+        public byte InnateIndex { get; set; }
+
+        public void Read(GamePacketReader reader)
```
Rawaho: Relating to the comment above, unless a packet is bidirectional it shouldn't inherit both IReadable and IWritable.
If the packet is in fact bidirectional name it appropriately.

Source/NexusForever.WorldServer/Network/Message/Model/ServerPlayerInnate.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            InnateIndex = reader.ReadByte(8u);
+        }
+
+        public void Write(GamePacketWriter writer)
+        {
+            writer.Write(InnateIndex, 2u);
```
Rawaho, line 14: Why is the innate index 8 bits in one packet and 2 bits in another?
Are you sure they are the same value?

Rawaho (changes requested):

Source/NexusForever.WorldServer/Network/Message/Model/ServerPlayerInnate.cs
kirmmin: It's not; I'll make this a Shared model.

Source/NexusForever.WorldServer/Network/Message/Model/ClientInnateChange.cs
kirmmin: Yep. Sorry boss.

Source/NexusForever.WorldServer/Network/Message/Model/ServerPlayerInnate.cs
kirmmin: I am positive :) I confirmed in my sniffs - I also thought it was odd, but then this might be one of the first packets they wrote and they never bothered updating it?

Rawaho: My concern is it might be similar to the ServerCombatReward situation where multiple types of values are packed into the same field.
Looking at sniffs alone might not reveal this, having a look at the code in the client that sends and receives these packets would give more insight into if this was the case or not.

kirmmin: Gotcha; To add, the only LUA events I had firing, when I received the packet was the Stance Update one.
But, I will confirm in the client code :)

kirmmin: Updated packet handler inheritance.

Additionally, dug into the packet being passed to and from the client. The server packet is only handled by the StanceChange LUA event, and the client packet doesn't seem to be used for anything else that I can see.

kirmmin:
Dug into the packet being passed to and from the client. The server packet only fires the `StanceChange` LUA event, and the client packet doesn't seem to be used for anything else that I can see.

PR #175, opened 2019-06-06
Feature: Account Currency support

kirmmin:


Source/NexusForever.WorldServer/Game/Account/AccountCurrencyManager.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// </summary>
+        public bool CanAfford(AccountCurrencyType currencyType, ulong amount)
+        {
+            return CanAfford((byte)currencyType, amount);
+        }
+
+        /// <summary>
+        /// Returns whether the Account has enough of the currency to afford the amount
+        /// </summary>
+        public bool CanAfford(byte currencyId, ulong amount)
```
Rawaho: Using AccountCurrencyType everywhere should be encouraged, having a duplicate method that take in a byte prevents this.
Anywhere that might need to use a byte (should be very limited) can just cast to AccountCurrencyType before calling the function.

Source/NexusForever.WorldServer/Game/Account/AccountCurrencyManager.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// </summary>
+        public void CurrencyAddAmount(AccountCurrencyType currencyType, ulong amount, ulong reason = 0)
+        {
+            CurrencyAddAmount((byte)currencyType, amount);
+        }
+
+        /// <summary>
+        /// Add a supplied amount to an <see cref="AccountCurrency"/>.
+        /// </summary>
+        public void CurrencyAddAmount(byte currencyId, ulong amount, ulong reason = 0)
```
Rawaho: Same as aboved.

Source/NexusForever.WorldServer/Game/Account/AccountCurrencyManager.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// </summary>
+        public void CurrencySubtractAmount(AccountCurrencyType currencyType, ulong amount, ulong reason = 0)
+        {
+            CurrencySubtractAmount((byte)currencyType, amount);
+        }
+
+        /// <summary>
+        /// Subtract a supplied amount to an <see cref="AccountCurrency"/>.
+        /// </summary>
+        public void CurrencySubtractAmount(byte currencyId, ulong amount, ulong reason = 0)
```
Rawaho: Same as above.

Source/NexusForever.WorldServer/Game/Account/AccountCurrencyManager.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            if (!accountCurrency.CanAfford(amount))
+                throw new ArgumentException($"Trying to remove more currency {accountCurrency.CurrencyId} than the player has!");
+
+            // TODO: Ensure that we're not at cap - is there a cap?
+            if(accountCurrency.SubtractAmount(amount))
+                SendAccountCurrencyUpdate(accountCurrency, reason);
+        }
+
+        private IEnumerable<ServerAccountCurrency> GetAccountCurrencyPackets()
```
Rawaho: You could merge this with the SendCharacterListPacket function using something like:
```
AccountCurrencies = currencies.Values.Select(c => c.BuildServerPacket()).ToList()
```
Or build them in a loop inside of the SendCharacterListPacket function, either way this function isn't required.

Source/NexusForever.WorldServer/Game/Account/AccountCurrency.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            saveMask = AccountCurrencySaveMask.None;
+        }
+
+        public bool AddAmount(ulong amount)
+        {
+            Amount += amount;
+            saveMask |= AccountCurrencySaveMask.Amount;
+
+            return true;
```
Rawaho, line 76: Is there even a reason to return a boolean if the result will always be true?

Source/NexusForever.WorldServer/Game/Account/AccountCurrency.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            Amount -= amount;
+            saveMask |= AccountCurrencySaveMask.Amount;
+
+            return true;
+        }
+
+        public bool CanAfford(ulong amount)
+        {
+            if (amount == 0)
```
Rawaho: Checking for 0 isn't needed since the check below will fail anyway if the amount is 0.

Source/NexusForever.WorldServer/Game/Account/AccountCurrencyManager.cs
```diff
@@ -0,0 +1,164 @@
...
+        public void Save(AuthContext context)
+        {
+            foreach (AccountCurrency accountCurrency in currencies.Values)
+                accountCurrency.Save(context);
+        }
+
+        /// <summary>
+        /// Create a new <see cref="CharacterCurrency"/>.
+        /// </summary>
+        private AccountCurrency CreateAccountCurrency(byte currencyId, ulong amount = 0)
```
Rawaho: Same as below.

Source/NexusForever.WorldServer/Game/Account/AccountCurrencyManager.cs
```diff
@@ -0,0 +1,164 @@
...
+using System;
+using NexusForever.Shared.GameTable;
+using NexusForever.Shared.GameTable.Model;
+using NexusForever.WorldServer.Game.Account.Static;
+
+namespace NexusForever.WorldServer.Game.Account
+{
+    public class AccountCurrencyManager
+    {
+        private static readonly ILogger log = LogManager.GetCurrentClassLogger();
```
Rawaho: No logging actually happens so this can be removed?

Rawaho (changes requested):

Source/NexusForever.WorldServer/Game/Account/AccountCurrency.cs
kirmmin: I left this as a bool in case it's capped (I just haven't added all the capping logic). For example, Omnibits have a weekly cap.

kirmmin:
Everything but 1 comment should be resolved, now. Thanks for the review :)

PR #179, opened 2019-06-08
Entity/Properties: Property updating ground work

kirmmin:
Big thanks to @jasoncouture who originally did the work for this :)

- Base properties are added based on level. There are only a handful, and this is only taken from a new character's perspective. If player's gained extra properties after a certain level, they are not added currently.
- Item Innate Properties are applied to the Player
   - This is only for Armor, HP, Attack Power and Support Power.
   - This does not include any secondary properties (like crit chance, vigor and deflect chance).
- Spells will now update properties like movement speed
   - This works for Mounts and spells like Sprint
   - This has been tested with spells that increase properties, but the effects of the property changes are not handled right now.
- Barebones implementation of stat updating. A single timer that runs every 0.25s is currently utilised.
   - The health stat will regenerate to maximum health value at roughly the rate of live (2% per sec).
   - The shield stat will regenerate to maximum shield value at roughly the rate of a standard shield (15% per second).
   - There are some TODOs added to account for the need to use different timers on different stat updates.


kirmmin:
I'm going to be making some updates to this PR to include base properties based on my findings. They will be modifiable, but they will be generated from sniffed data.

Source/NexusForever.WorldServer/Game/Entity/WorldEntity.cs
```diff
@@ -135,16 +145,46 @@ public virtual void OnActivateCast(Player activator)
         {
             // deliberately empty
         }
+        
+        public virtual ServerEntityPropertiesUpdate BuildPropertyUpdates()
+        {
+            if (!HasPendingPropertyChanges)
+                return null;
+            HashSet<Property> properties = new HashSet<Property>();
```
Therzok: This isn't used?

kirmmin: Yep, a bad copy-paste from the reference code I had. I moved this to higher scope so that it could be modified appropriately.
I'll get it updated ;) Thanks.

kirmmin:
Updated this PR following recent research. This enables storing of Property values depending on their source, and also removal (Spell Removal is not implemented at this time). Property values update using accurate data derived from the spell tables, currently converted using `BitConverter`.
I have not added the layout update for the GameTable yet.

kirmmin:
I've finalised the Property work for this initial pass. It will not be the final implementation of the Property updating system, and specifically Item Properties, but this PR gives the following:

- Base properties are added based on level. There are only a handful, and this is only taken from a new character's perspective. If player's gained extra properties after a certain level, they are not added currently.
- Item Innate Properties are applied to the Player
   - This is only for Armor, HP, Attack Power and Support Power.
   - This does not include any secondary properties (like crit chance, vigor and deflect chance).
- Spells will now update properties like movement speed
   - This works for Mounts and spells like Sprint
   - This has been tested with spells that increase properties, but the effects of the property changes are not handled right now.
- Barebones implementation of stat updating. A single timer that runs every 0.25s is currently utilised.
   - The health stat will regenerate to maximum health value at roughly the rate of live (2% per sec).
   - The shield stat will regenerate to maximum shield value at roughly the rate of a standard shield (15% per second).
   - There are some TODOs added to account for the need to use different timers on different stat updates.

kirmmin:
Rebased this PR to include the new Singleton updates.

I also discovered that this PR breaks Rapid Transport. I believe a property is being assigned and that's causing the client to say `Unable to recall while in Unhealthy Time`. Needs investigation so I'm placing this back in WIP.

kirmmin:
The issue was resolved by adding the PreRequisite cast support for Recall spells. I'm not 100% sure what the error message alludes to, but I assume it was Health not being full, and added the PreReq support with that.

WIP removed and ready to merge again

kirmmin:
Rebased and updated today. Now part of DB migrations.

Issue #180, opened 2019-06-10
[Bug] - [Character Creation]: Name Dupe


HelperDave:
**TITLE**: 
Naming Dupe

**DATE**: 
10 June 2019

**DESCRIPTION**:
You can currently create a character with a name already taken.

Initially you won't be able to play this character upon creation; (cinematic freeze) however restarting the client will allow you to play the character.

**OCCURRENCE**:
Always

**OS**: 
Win10 - Latest

**BRANCH/COMMIT**:
Develop (latest?) - (Kirmmins)

**STR**: 
Create a character with the same name of another character.

**OBSERVED**:
Initially you won't be able to play this character upon creation; (cinematic freeze) however restarting the client will allow you to play the character.

**EXPECTED**:
Name taken feature implemented.
Certainly to not play the character even if created!

![WildStar 190610 175616](https://user-images.githubusercontent.com/51673399/59211952-18018f00-8ba9-11e9-9a81-a958a5d256e0.png)

dyl10s:
This issue can be closed. @Rawaho

PR #184, opened 2019-06-17
Bug Fix: Check that a new character has a unique name

dyl10s:
This is my first pull request. I tried to take on something simple. I am very open to feedback on this. My changes fix issue #180 with a generic error if the name already exists


Source/NexusForever.WorldServer/Database/Character/CharacterDatabase.cs
```diff
@@ -87,6 +87,14 @@ public static async Task<List<Model.Character>> GetCharacters(uint accountId)
             }
         }
 
+        public static bool CharacterNameExists(string characterName)
+        {
+            using (var context = new CharacterContext())
+            {
+                return context.Character.Where(c => c.Name == characterName).Count() == 0 ? false : true;
```
Fabi: If you do it like that, you can use  `context.Character.Any(c => c.Name == characterName);`

dyl10s: That seems more readable. This is entity framework correct? I have never used it before.

kirmmin:
Good job on the PR, @dyl10s . 

I'd encourage you to use appropriate result codes. They are available here: https://gist.github.com/kirmmin/1b632efc95abb791d0fdc070fe486f20. This file does not exist in the codebase at the moment, it's a dump from the `CodeEnumCharacterModifyResult` enum in the Client's code, so please include this file as part of your PR. I would also suggest updating the packet handler to use this Enum, too.

dyl10s:
> CodeEnumCharacterModifyResult

@kirmmin 
Is this what you meant by updating the packet handler?

kirmmin:
> > CodeEnumCharacterModifyResult
> 
> Is this what you meant by updating the packet handler?

Exactly right. Good job!

PR #185, opened 2019-06-17
Spells/Core: Telegraph additions (Circle, LongCone, Square, Rectangle)

kirmmin:
I wouldn't say this was hacked together, but it's not far off that :) Thank you @Rawaho for the reference code that I blatantly plagiarised.

I also took a quick look at the offset positions, they don't seem overly difficult to figure out, but I believe that the offset will need to happen at Telegraph instantiation rather than at the damage check. For now, I have it calculating "forward" offset in the telegraph check.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Spell/Static/DamageShape.cs
+++ b/Source/NexusForever.WorldServer/Game/Spell/Static/DamageShape.cs
@@ -2,7 +2,10 @@
 {
     public enum DamageShape
     {
-        Cone      = 4,
-        Rectangle = 7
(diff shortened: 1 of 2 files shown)


kirmmin:
Tidied this up and made some optimisations. It's a lot cleaner now. Looking for your review please, time permtting, @Rawaho ;)

This will complete some of the portions of #308

xNxExOx (commented):
I did not verify corectness of the enum values, but the intersection math seems correct.
(P.S. I personally do not like double indentation of the `case` code, and it is inconsistent with previous code)

PR #197, opened 2019-08-15
Remodels and Decor Linking

senna-fox:
Allows the server to handle, store, and distribute information regarding Housing Remodels and Decor Linking. Also provides a better fix to the issue of decor offsets, though a another
solution may need to be found in the future.


Source/NexusForever.WorldServer/Game/Housing/Residence.cs
```diff
@@ -106,6 +106,25 @@ public ushort Door
...
+        public ushort Music
+        {
+            get => musicId;
+            set
+            {
+                HousingWallpaperInfoEntry entry = GameTableManager.HousingWallpaperInfo.GetEntry(value);
+                if (entry == null)
+                    throw new ArgumentOutOfRangeException();
+
+                if ((entry.Flags & 0x180) == 0)
```
Rawaho: 0x0180 is 2 flags, 0x0080 and 0x0100, I would say only one of these actually determines if it is music or not.

Source/NexusForever.WorldServer/Game/Map/ResidenceMap.cs
```diff
@@ -279,7 +281,13 @@ private void DecorCreate(Player player, ClientHousingDecorUpdate.DecorUpdate upd
                     throw new InvalidPacketValueException();
 
                 // new decor is being placed directly in the world
-                decor.Position = update.Position;
+                var position = new Vector3(update.Position.X, update.Position.Y, update.Position.Z);
+                if (update.PlotIndex != 0)
+                {
+                    position.Y += 0.835f;
+                    position.Z += 0.01f;
+                }
```
Rawaho: A nitpick but I might save to have modifying the offset in a nested function rather than in multiple places.
I would say eventually we will need one anyway once we understand the offsets some more.

Source/NexusForever.WorldServer/Game/Map/ResidenceMap.cs
```diff
@@ -311,6 +319,13 @@ private void DecorMove(ClientHousingDecorUpdate.DecorUpdate update)
                 throw new InvalidPacketValueException();
 
             // TODO: research 0.835f
+            // in meantime workaround checks for House PlotIndex and sets hardcoded offset
+            var position = new Vector3(update.Position.X, update.Position.Y, update.Position.Z);
+            if (update.PlotIndex != 0)
+            {
+                position.Y += 0.835f;
+                position.Z += 0.01f;
+            }
```
Rawaho: As above.

Source/NexusForever.WorldServer/Network/Message/Model/ServerHousingProperties.cs
```diff
@@ -45,13 +45,13 @@ public void Write(GamePacketWriter writer)
...
-                writer.Write(0);
-                writer.Write(0);
-                writer.Write(Ground);
+                writer.Write(Roof);
+                writer.Write(WallpaperExterior);
+                writer.Write(Entryway);
+                writer.Write(Door);
                 writer.Write(Sky);
+                writer.Write(Music);
+                writer.Write(Ground);
```
Rawaho, line 54: Since an additional value has been added here do the values below still match?

Rawaho (changes requested):
Overall a great PR, only a few minor issues and queries.

Source/NexusForever.WorldServer/Network/Message/Model/ServerHousingProperties.cs
senna-fox: As best as I know the values below are still good. Nothing was actually added here, it was simply given a value to write in, and arranged so that the packet would be written correctly. Previously values were being written in no particular order and that was causing the client to misassign that data. i.e. sky data was being written in as grounds, etc.

Source/NexusForever.WorldServer/Game/Housing/Residence.cs
myyanf: Oddly, the Flags setting for music in the HousingWallpaperInfo entries is 0x180. Dumping flag values for entries 1-4000  from the list, I see that roughly the first 285 entries (minus a couple) exist, and that other than music with its 0x180 setting, nothing else appears to use either flag. Given this, technically either flag alone would work to determine if the entry is music,

PR #201, opened 2019-09-15
Adding the initial /roll functionality

myyanf:
This tracks the develop branch at the moment,
and as such, it does not include party and raid
group responses yet. /roll output will only be
returned to the user using /roll at this time.


Source/NexusForever.WorldServer/Network/Message/Model/ClientRandomRollRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    [Message(GameMessageOpcode.ClientRandomRollRequest)]
+    public class ClientRandomRollRequest : IReadable
+    {
+        // Note that for /roll on the command line, the first two fields below,
+        // realmId, characterId, and Unknown0-3 are 0.
+        // As such, these fields are currently unused.
+        public ushort RealmId { get; set; }
+        public ulong CharacterId { get; set; }
```
Rawaho: Use the TargetPlayerIdentity class instead of 2 separate fields.

Source/NexusForever.WorldServer/Network/Message/Model/ServerRandomRollResponse.cs
```diff
@@ -0,0 +1,24 @@
+﻿using NexusForever.Shared.Network;
+using NexusForever.Shared.Network.Message;
+
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    [Message(GameMessageOpcode.ServerRandomRollResponse)]
+    public class ServerRandomRollResponse : IWritable
+    {
+        public ushort RealmId { set; get; }
+        public ulong CharacterId { set; get; }
```
Rawaho: Same as above.

Source/NexusForever.WorldServer/Network/Message/Model/ClientRandomRollRequest.cs
```diff
@@ -0,0 +1,29 @@
...
+        public int MinRandom { get; set; }
+        public int MaxRandom { get; set; }
+        public int Unknown0 { get; set; }
+        public Random rnd = new Random();
```
Rawaho: This shouldn't be in the model, move it to the message handler function.

Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
```diff
@@ -61,5 +61,18 @@ public static void HandleWeaponToggle(WorldSession session, ClientToggleWeapons
         {
             session.Player.Sheathed = toggleWeapons.ToggleState;
         }
+
+        [MessageHandler(GameMessageOpcode.ClientRandomRollRequest)]
+        public static void HandleRandomRoll(WorldSession session, ClientRandomRollRequest randomRoll)
+        {
```
Rawaho, line 68: You need to do the same error checks the client does before sending the response.
The client does 2 main checks in the function that sends the 0x071B packet.
Is min larger than max?
Is max larger than 1000000?

Rawaho (changes requested):

Source/NexusForever.WorldServer/Network/Message/Model/ClientRandomRollRequest.cs
myyanf: Thanks. This is done.

Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
myyanf: Sorry for leaving it out. I know better. This has been fixed.
Also switched MinRandom and MaxRandom to uint to
avoid the need for <0 checks.

PR #202, opened 2019-09-21
Added the command to change the displayinfo.

crazysheep86:
Using the command "!DisplayInfo id (creatureID)" uses the command SetDisplayInfo on the player. Also checks to make sure its a valid creatureID to be able to turn into.


MaxtorCoder:
Also please a different branch, since you are merging Master into Dev.

Source/NexusForever.WorldServer/Command/Handler/DisplayInfoCommandHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+                return Task.CompletedTask;
+            }
+               
+            if (!uint.TryParse(parameters[0], out uint displayInfo))
+            {
+                context.SendMessageAsync("Parameters are invalid. Please try again.");
+                return Task.CompletedTask;
+            }
+
+            if (GameTableManager.Creature2.GetEntry(displayInfo) == null)
```
kirmmin: Pretty sure Creature2 is keyed by ID, not displayInfo. If you plan on having them enter creature IDs, then have the player mimic that creature, you'll need to feed in the displayInfo id from the Creature2 entry into the `SetDisplayInfo` method below.

kirmmin:
Please ensure you're rebasing PRs and not merging code in. Committed as is would result in commit bloat that's unnecessary

crazysheep86:
> Also please a different branch, since you are merging Master into Dev.

How do I change this? Should I make a whole new PR as it only allows me to change the Nexus branch from Dev or Master? I made a new branch on my fork, but I can't find a way to end that side on here.

Source/NexusForever.WorldServer/Command/Handler/DisplayInfoCommandHandler.cs
crazysheep86: I just wanted a check to make sure if its a valid number. Couldn't find a displayInfo on the creature2 gametable so thought the id's must be the same. I obviously thought wrong as I just tested the creature id's from the database instead of the gametable and see the error.

crazysheep86: is there a way to check the database to make sure displayInfo id is valid, or should I just cut that if statement out?

Rawaho: Creature2DisplayInfo has display id by key

crazysheep86: Now I feel dumb lol. Adding that change now

Source/NexusForever.WorldServer/Command/Handler/ModifyCommandHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+            : base(true, "modify")
+        {
+        }
+
+        [SubCommandHandler("displayInfo", "id - Change your look to match that of a creature.")]
+        public Task DisplayInfoSubCommandHandler(CommandContext context, string command, string[] parameters)
+        {
+            if (parameters.Length != 1)
+            {
+                context.SendMessageAsync("Parameters are invalid. Please try again.");
```
Rawaho: I would use `SendHelpAsync` instead of sending this error, it will send the help text to the player.
Also you should `await` any asynchronous calls.

Source/NexusForever.WorldServer/Command/Handler/ModifyCommandHandler.cs
```diff
@@ -0,0 +1,41 @@
...
+                return Task.CompletedTask;
+            }
+               
+            if (!uint.TryParse(parameters[0], out uint displayInfo))
+            {
+                context.SendMessageAsync("Parameters are invalid. Please try again.");
```
Rawaho: Same as above.

Rawaho (changes requested):

Source/NexusForever.WorldServer/Command/Handler/ModifyCommandHandler.cs
crazysheep86: Ok I'll get those changes done, I copied it basically from the currency Command, might need to update those too

PR #204, opened 2019-09-22
Fix: Rotation applies when pathing

kirmmin:


Rawaho:
I'm going to close this as it doesn't work for any path that isn't straight.
What needs to happen is the rotation between each node needs to be calculated and send with the spline time offset to the client in bulk.
We need to handle entity times properly before this can be added.

PR #207, opened 2019-09-28
Housing/Fix: Decor offset resolved

kirmmin:


kirmmin:
PR has been updated with validation of `PlotIndex` parameter passed from the client. `HousingResult` handler was also added to allow for simple implementation of Decor placement validation.

PR #213, opened 2019-10-11
Spells/Core: Spell Casting MEGAUPDATE

kirmmin:
After some time spent working on spell casting improvements, this is now at a place where it really needs a review.
I've made heavy use of the spell event system to provide correctly timed spell casting. This may not be the most efficient resource usage, so a review on this part in particular is requested. A lot of time was spent researching the spell casting packets and how they instruct the client what to do. From what I can tell, our spell packets trigger and mimic retail's, now (minus unknown packets).

- Cast methods added
   - Multiphase (Used very frequently by NPC spells)
   - Rapid Tap
   - Charge and Release
- Spell finishing
   - This update includes the functionality to end spell casts at the appropriate duration.
- Buff Timers
   - Added buff timers in line with how they worked in spell packets from retail.
- Proxy Spell fixes
   - Proxy spells that ticked at a period interval were presented differently in the tables. The proxy handler has been corrected to handle this appropriately.
- Threshold Casting
   - These are, essentially, steps to a spell cast that require 2 or more button presses. This is used as part of Rapid Tap and Charge and Release casting.
- Phased Casts
   - These phases indicate trigger points during a single spell cast. Spells like Shred which hit 3 times during a single cast make use of spell phases. This is used as part of Multiphase casting.
- ~~`UnlockedSpell` handler~~
   - ~~All LAS spell casts now take place "through" the `UnlockedSpell` instance associated with the button. This enables the server to better keep context of an existing spell, and opens up possibilities for improvements later (like tracking Spell Cooldown via the `UnlockedSpell` instance so we don't use server resources on re-checking every time an ability cast is requested)~~
   - Replaced by `CharacterSpell` that got added as part of another PR.

### DIFF
--- a/Source/NexusForever.Shared/GameTable/GameTableManager.cs
+++ b/Source/NexusForever.Shared/GameTable/GameTableManager.cs
@@ -536,7 +536,9 @@ namespace NexusForever.Shared.GameTable
         [GameData]
         public GameTable<Spell4TelegraphEntry> Spell4Telegraph { get; private set; }
 
+        [GameData]
         public GameTable<Spell4ThresholdsEntry> Spell4Thresholds { get; private set; }
+
         public GameTable<Spell4TierRequirementsEntry> Spell4TierRequirements { get; private set; }
 
         [GameData]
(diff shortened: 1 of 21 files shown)


kirmmin:
Replaced by [spells-v2 branch](/NexusForever/NexusForever/tree/spells-v2)

PR #219, opened 2019-11-07
Core/RealmList: Reload servers and handle offline

kirmmin:
Refreshes servers from database if the server list is deemed dirty
This will kick the user back to login if they try to join a server that is marked as offline. It's not elegant at this time, but it works.

Note: This was written so that Admins could add/remove extra realms from their Auth table without having to bump any existing World Servers.

### DIFF
--- a/Source/NexusForever.Shared/Game/ServerInfo.cs
+++ b/Source/NexusForever.Shared/Game/ServerInfo.cs
@@ -14,7 +14,7 @@ namespace NexusForever.Shared.Game
 
         public Server Model { get; }
         public uint Address { get; }
-        public bool IsOnline { get; private set; } = false;
+        public bool IsOnline { get; private set; }
 
         public ServerInfo(Server model)
         {
@@ -38,20 +38,17 @@ namespace NexusForever.Shared.Game
             }
         }
 
-        public async Task PingHost(string hostIP, int port)
+        /// <summary>
+        /// Attempt to connect to the remote world server asynchronously.
+        /// </summary>
+        public async Task PingHostAsync()
         {
-            using (var client = new TcpClient())
+            using var client = new TcpClient();
+            await client.ConnectAsync(Model.Host, Model.Port).ContinueWith(task =>
             {
-                await client.ConnectAsync(hostIP, port).ContinueWith(task =>
-                {
-                    if (task.IsFaulted)
-                        IsOnline = false;
-                    else
-                        IsOnline = true;
-
-                    return Task.CompletedTask;
-                });
-            }
+                IsOnline = !task.IsFaulted;
+                return task;
+            });
         }
     }
 }
--- a/Source/NexusForever.Shared/Game/ServerInfo.cs
+++ b/Source/NexusForever.Shared/Game/ServerInfo.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Net;
 using System.Net.Sockets;
+using System.Threading.Tasks;
 using NexusForever.Shared.Database.Auth.Model;
 using NLog;
 
@@ -13,6 +14,7 @@ namespace NexusForever.Shared.Game
 
         public Server Model { get; }
         public uint Address { get; }
+        public bool IsOnline { get; private set; } = false;
 
         public ServerInfo(Server model)
         {
@@ -33,6 +35,22 @@ namespace NexusForever.Shared.Game
             {
                 log.Fatal(e, $"Failed to load server entry id: {model.Id}, host: {model.Host} from the database!");
                 throw;
+            }
+        }
+
+        public async Task PingHost(string hostIP, int port)
+        {
+            using (var client = new TcpClient())
+            {
+                await client.ConnectAsync(hostIP, port).ContinueWith(task =>
+                {
+                    if (task.IsFaulted)
+                        IsOnline = false;
+                    else
+                        IsOnline = true;
+
+                    return Task.CompletedTask;
+                });
             }
         }
     }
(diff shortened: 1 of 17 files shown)


Rawaho:
The problem with this implementation is the updating of the servers happens on the world thread, this will block updating the next session while long running take such as synchronous database calls and TCP connections happen.
You really want all of that to happen in another thread.

kirmmin:
Thanks @Rawaho. Been a while since I've done a database lookup during a thread so I had forgotten this needs to run asynchronously when making database calls. I'll update appropriately :)

kirmmin:
@Rawaho, I've updated the code following your review. I'm still not 100% sure how await/async works in C#-world, so a follow up review will really help <3

Rawaho:
GetResult synchronously waits for an asynchronous task to complete which defeats the point in this scenario.

For the database calls you don't need to wrap the entire thing in a task, EF already has asynchronous methods to get data from the database.
Converting to ImmutableList afterwards can be synchronous as it will be quick compared to the database call.

You are also still pinging the world server on the world thread, this means that while the world server is trying to establish the connection to the other world server the thread is "paused" and no session or world updates can occur.

My suggestion would be to have either a long running task or separate thread continuously updating the server status and when you try and connect it just checks the model which will have an up to date status.

kirmmin:
Okay! I was tempted to go that way, but not sure how you felt about running an Update method on Server Manager, so opted not to. Sounds like it would've been a better way to go though :)
I'll refactor again, and hopefully get it right this time ;)

Thanks a lot!

kirmmin:
@Rawaho, I hope this'll be the last review you need to do on this. I've moved it to an Update loop and made the Ping check Asynchronous (I think). Please confirm I'm going in the right direction, when you have time, of course.

kirmmin:
Updated ServerManager to use its own thread as discussed on Discord. While playing with a verification method, I also updated the `ServerHello` packet structure to have more named parameters.

This isn't the final form of the online status check, but this should be a good first start (assuming no further changes). :)

Source/NexusForever.Shared/Game/ServerManager.cs
```diff
@@ -9,6 +13,11 @@ public sealed class ServerManager : Singleton<ServerManager>
         public ImmutableList<ServerInfo> Servers { get; private set; }
         public ImmutableList<ServerMessageInfo> ServerMessages { get; private set; }
 
+        private double CheckDuration = 60d; // TODO: Make this configurable?
+        private double checkTimer;
```
Rawaho: Could be replaced with UpdateTimer.

Source/NexusForever.Shared/Game/ServerInfo.cs
```diff
@@ -28,12 +32,43 @@ public ServerInfo(Server model)
...
         }
+
+        public void Update(double lastTick)
+        {
+            pingTimer -= lastTick;
+            
+            if(pingTimer <= 0d)
+            {
+                // check server
+                PingHost(Model.Host, Model.Port);
```
Rawaho: Asynchronous function isn't awaited, could lead to thread safety issues.

Source/NexusForever.Shared/Game/ServerInfo.cs
```diff
@@ -28,12 +32,43 @@ public ServerInfo(Server model)
...
+
+                pingTimer = PingDuration;
+            }
+        }
+
+        public async void PingHost(string hostIP, int port)
+        {
+            using (var client = new TcpClient())
+            {
+                await client.ConnectAsync(hostIP, port).ContinueWith(task =>
```
Rawaho: This probably doesn't need to be asynchronous as it is handled in a separate thread.

Source/NexusForever.Shared/Game/ServerInfo.cs
```diff
@@ -7,12 +7,16 @@
...
+    public class ServerInfo: IUpdate
     {
         private static readonly ILogger log = LogManager.GetCurrentClassLogger();
 
         public Server Model { get; }
         public uint Address { get; }
+        public bool IsOnline { get; private set; } = false;
+
+        private double PingDuration = 15d;
+        private double pingTimer;
```
Rawaho: Could be replaced with UpdateTImer.

Source/NexusForever.Shared/Game/ServerInfo.cs
```diff
@@ -28,12 +32,43 @@ public ServerInfo(Server model)
 
                 Address = (uint)IPAddress.HostToNetworkOrder(BitConverter.ToInt32(ipAddress.GetAddressBytes()));
                 Model   = model;
+
+                PingHost(Model.Host, Model.Port);
```
Rawaho: Asynchronous function isn't awaited, could lead to thread safety issues.

Rawaho (changes requested):

Source/NexusForever.Shared/Game/ServerInfo.cs
Fabi: Unlikely. async functions with no Task as return type are not supposed to be awaited but will just run and not block the current method. Should be fine

Source/NexusForever.Shared/Game/ServerInfo.cs
kirmmin: I've updated everything except this. I decided to keep it asynchronous because it was cleaner code but also meant that we didn't need to wait for a connection to 1 server to end before moving on to check the next, we can check them all in parallel, or near enough.

PR #223, opened 2019-11-13
SocialManager/Core: Fix for chat handler delegates

### LINKED ISSUE #222: [Bug] Social/Core: Chat Handler delegation now fails
Date: November 13th, 2019

Description: It appears that Chat Handlers are now not being Initialised properly when the SocialManager class is loaded.

Occurrence: All.

OS: Windows 10

Branch/Commit: develop

STR:

1. Boot up a server with a fresh develop branch
2. Login to the game
3. Try and write things in /say, /emote, or /yell.

Observed: The Debug text comes back saying "SocialManager: Currently not implemented" for each time you try to "say" something.
The emote commands like `/laugh` and such still work

Expected: The debug text does not come back and your message goes through.

kirmmin:
Fixes #222

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Social/SocialManager.cs
+++ b/Source/NexusForever.WorldServer/Game/Social/SocialManager.cs
@@ -43,11 +43,8 @@ namespace NexusForever.WorldServer.Game.Social
 
         private void InitialiseChatHandlers()
         {
-            IEnumerable<MethodInfo> methods = Assembly.GetExecutingAssembly()
-                .GetTypes()
-                .SelectMany(t => t.GetMethods(BindingFlags.NonPublic | BindingFlags.Static));
-
-            foreach (MethodInfo method in methods)
+            foreach (MethodInfo method in Assembly.GetExecutingAssembly().GetTypes()
+                .SelectMany(t => t.GetMethods(BindingFlags.NonPublic | BindingFlags.Instance)))
             {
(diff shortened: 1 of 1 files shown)


Source/NexusForever.WorldServer/Game/Social/SocialManager.cs
```diff
@@ -59,8 +59,15 @@ private void InitialiseChatHandlers()
                     Debug.Assert(typeof(ClientChat) == parameterInfo[1].ParameterType);
                     #endregion
 
-                    ChatChannelHandler @delegate = (ChatChannelHandler)Delegate.CreateDelegate(typeof(ChatChannelHandler), method);
-                    chatChannelHandlers.Add(attribute.ChatChannel, @delegate);
+                    var c1 = Expression.Parameter(typeof(WorldSession));
```
Rawaho: You shouldn't need to generate a delegate from an expression tree, CreateDelegate supports both static and instance methods.

kirmmin: I thought the same thing, but I was getting errors trying to use `CreateDelegate`. I'll take another look later.

kirmmin (commented):
Converted back to Delegate. Issue was that I wasn't giving the `CreateDelegate` method the correct instance to bind the method against :)

This is the method I needed to use, for future reference: https://docs.microsoft.com/en-us/dotnet/api/system.delegate.createdelegate?view=netframework-4.8#System_Delegate_CreateDelegate_System_Type_System_Object_System_Reflection_MethodInfo_

Issue #215, opened 2019-10-17
Decor links remain when decor is crated


kirmmin:
**Date:** October 16th, 2019

**Description:** As the title says, when you crate decor the links remain, so when you re-place the Decor, items are placed "still inked" which may have funky side effects.

**Occurrence:** Always

**OS:** Windows 10

**Branch/Commit:** Latest dev branch

**STR:** 
1. Place 2 or more decor
2. Link the decor
3. Crate the decor
4. Place them both in different positions
5. Notice the link

**Observed:** Decor remains linked when placing again and moving the parent Decor will move the children

**Expected:** Decor should be unlinked when placing again

PR #225, opened 2019-11-14
Fix for decor not unlinking when crated. Resolves issue #215.

myyanf:


myyanf:
After a couple hours of sleep, I realized that the fix for crating the parent decor item was likely just a caching issue. Verified that this morning. I'll work on the fix for crating parents as soon as possible.

This simple patch fixes the issue for any crated children decor items at the moment. -Myya

myyanf:
When crating the parent decor, a ClientHousingDecorUpdate packet is sent with updates for all of the children, setting the ParentDecorId to 0, so with the corrected file submitted, this corrects #215 .

PR #231, opened 2019-11-29
Spells/ActionSet: Fixed Save Issues, Added Macro support

kirmmin:
Updated April 13th, 2020.
- Fixed Saving of Shortcuts and AMPs. Should also fix load issues related to AMPs
- Fixed removal of shortcuts from the non-LAS hotbars.
- Added Macro support. Macros are not stored server side at this time, but Macro shortcuts should be stored between sessions.

---

Main problem previously was that Amps and Shortcuts that were looking to be created could've tried to be created earlier than the existing one was being deleted. This forces the deleted shortcuts and amps to execute first "freeing up" the entry's keys to be used by the newly created entry.

### DIFF
--- a/Source/NexusForever.Database.Character/CharacterContext.cs
+++ b/Source/NexusForever.Database.Character/CharacterContext.cs
@@ -243,7 +243,8 @@ namespace NexusForever.Database.Character
                 entity.Property(e => e.SpecIndex)
                     .HasColumnName("specIndex")
                     .HasColumnType("tinyint(3) unsigned")
-                    .HasDefaultValue(0);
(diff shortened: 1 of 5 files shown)


kirmmin:
Marked as WIP because I'm going to refactor this slightly

kirmmin:
Refactored and rebased today.

- Fixed Saving of Shortcuts and AMPs. Should also fix load issues related to AMPs
- Fixed removal of shortcuts from the non-LAS hotbars.
- Added Macro support. Macros are not stored server side at this time, but Macro shortcuts should be stored between sessions.

PR #236, opened 2019-12-04
Core/Movement: Fix client warping

kirmmin:
Following a conversation in #dev_general channel, it turns out that the Client Entity Command "Echos" should not have the `ServerControlled` flag set to `true`. This was confirmed in packet logs of other players Entity Commands being sent to the client.
This greatly changed the way in which 1 client saw another client. Previously, there was a lot of warping and "missed frames" from player to player, especially when both were a distance from a server, say 50ms ping. With this change, the warping had decreased greatly due to the Client handling the command differently.

Example client entity command packet received after another player's movement:
```
[Server [0x0638] ServerEntityCommand]
{
    "GUID": 2974686,
    "Time": 115796415,
    "TimeReset": false,
    "ServerControlled": true,
    "commands": [
        {
            "Command": "SetTime",
            "Data": {
                "time": 115796539
            }
        },
        {
            "Command": "SetPosition",
            "Data": {
                "x": -3234.896728515625,
                "y": -907.4629516601562,
                "z": -765.501953125
            }
        },
        {
            "Command": "SetVelocity",
            "Data": {
                "x": 50932,
                "y": 48822,
                "z": 18659
            }
        },
        {
            "Command": "SetMove",
            "Data": {
                "x": 50908,
                "y": 0,
                "z": 18668
            }
        },
        {
            "Command": "SetRotation",
            "Data": {
                "x": 2.533184289932251,
                "y": 0,
                "z": 0
            }
        }
    ]
}
```

### DIFF
--- a/Source/NexusForever.WorldServer/Network/Message/Handler/EntityHandler.cs
+++ b/Source/NexusForever.WorldServer/Network/Message/Handler/EntityHandler.cs
@@ -45,7 +45,7 @@ namespace NexusForever.WorldServer.Network.Message.Handler
             {
                 Guid     = mover.Guid,
                 Time     = entityCommand.Time,
-                ServerControlled = true,
+                ServerControlled = false,
                 Commands = entityCommand.Commands
             });
         }


kirmmin:
I opted to be explicit about this value being set to `false` and not just deleting it because if someone changes the packet's default in the future, it could break this change.

PR #245, opened 2019-12-10
Misc/Fix: Deleted character kick the client out

kirmmin:
Throwing the exception causes the client to be disconnected. This makes the client show this other player as "Unknown" in the Mail "from" field, friends' lists, guild lists', etc. 

This fix is mainly for Mail, as a player should be removed as a friend, removed from guild, removed from communities, and neighbours list, on deletion.


Rawaho:
This isn't an ideal fix as it's masking an actual problem.
What really should happen is CharacterManager should still track deleted characters with some field saying the returned ICharacter is actually deleted.

PR #247, opened 2019-12-11
Inventory/Fix: BagIndex dupe save issue

kirmmin:
Deleted Items need to be saved first to "free up" the `Location` and `BagIndex` for use by the new Item.


kirmmin:
After posting this, I realised that this wasn't necessarily an issue due to the fact that we key by GUID, but it still seems a safer bet to clean up first before we add new entries. So, even if it doesn't fix any issue, it's still better practice.

PR #249, opened 2020-01-04
Fix Create Acc with capital first letter

mifrgr:
Server sends email/username convert to lowercase (thanks Kirmmin) to the server, so the change will create a username/email withcomplete lower letters


Fabi:
What about creating triggers that auto lowercase on insert at sql server side?

mifrgr:
For me it seems the easiest way, with at less code changes as possible, so it would be easy to overview the pr and merge it. It works and it is similar to the way it works on the client ( where you can input capital letters and the client convert it to lowercase before it send it to the server) I'm pretty sure there are other ways to solve it, but I thought this has the most cost-effectiveness

Rawaho:
I tried to rebase this for you.
Was throwing permission denied, you'll need to do this yourself.

PR #257, opened 2020-04-02
Map/Zones: Zone Change Client packet

kirmmin:
I was tired of seeing `063A` pop up constantly and with no clue what it meant 😛 Now this makes sense. I haven't hooked any functionality up to it, but would be a good indicator when we want to apply scripts when a Player is in a certain area, like the gravity effect modifier for The Loftite Cliffs. Or, trigger a Challenge, for example.


Rawaho:
This packet was deliberately ignored and will probably stay that way.
While having the empty handler is good to prevent to spam in the console we shouldn't rely on it to trigger events as it comes from the client.
We already have a hook at the server that notifies when the player changes zone using the base maps.

kirmmin:
Absolutely agree with you :) I mainly wanted to stop the notifications, and was just a suggestion on how we could use Client calls to help in server development to confirm or deny certain assumptions (when in Debug mode, for example).
Feel free to decline this PR if you don't wish to apply it, just thought it would be good one to "figure out" so that the spam could somewhat be reduced 😄

PR #260, opened 2020-04-06
Spell/Effects: Stealth, Cooldown and ForceRemove - WIP

kirmmin:
This has a rudimentary implementation of "Status Effects" for a World Entity. I believe it can be expanded on without much work, but some pointers would help because I was a little lost about how best to implement this type of logic without spending many cycles figuring out what statuses to apply or remove.


kirmmin:
Did some further investigation of the Stealth passives. They all have Proc effects which trigger the ability to be removed when the Player casts abilities. Obviously, I didn't include Procs as part of this PR, but the WS Spell system relies on Procs _a lot_ to do this type of stuff, so it might be something that we'll want to add in sooner rather than later.

kirmmin:
Marking as WIP so I can implement Cooldown modifier types properly.

(This note is mainly for myself :P )

PR #261, opened 2020-04-11
Cinematics/Core: Initial implementation

kirmmin:
- Includes Novice Tutorial starting Cinematic for both Factions
- Includes Exile cinematic for the Combat step of the Novice Tutorial

I wrote this after finally getting some packets working from the Novice Tutorial zone. It's a complicated thing to write a system for, because there are many different cinematics, and requires a lot of parsing to get the information out - then manual entry. The way I've implemented this does lend itself to being able to generate scripts for it, though that might not be something we want to do in the short term as we test out different cinematic scenarios.
You can see the Exile one on this example I stuck on YouTube: https://youtu.be/sCqCJUFmcOU (Skip to around 1m35s).

This is something that will need iteration (and likely a bunch of tidying). I'd appreciate any reviews that would help clean this up.


Source/NexusForever.WorldServer/Game/Cinematic/GlobalCinematicManager.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace NexusForever.WorldServer.Game.Cinematic
+{
+    public class GlobalCinematicManager : Singleton<GlobalCinematicManager>
+    {
+        /// <summary>
+        /// Id to be assigned to the next Actor
+        /// </summary>
+        public uint NextCinematicId => nextCinematicId++;
+
+        private uint nextCinematicId = 1073743000;
```
Ness199X, line 12: What is this magic number?

Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
```diff
@@ -294,11 +294,11 @@ public static void HandleCharacterCreate(WorldSession session, ClientCharacterCr
                     });
                 }
 
-                //TODO: handle starting locations per race
-                character.LocationX = -7683.809f;
-                character.LocationY = -942.5914f;
-                character.LocationZ = -666.6343f;
-                character.WorldId = 870;
+                //TODO: handle starting locations per race - this is set to the Novice Tutorial zone
```
Ness199X: Might be a good opportunity to hook up different start locations based on novice/veteran start

PR #263, opened 2020-04-12
Implementation of Residence Flags settings.

myyanf:
Implements the flags for hiding ground clutter and neighbor skyplots.

### DIFF
--- a/Source/NexusForever.WorldServer/Network/Message/Model/ClientHousingFlagsUpdate.cs
+++ b/Source/NexusForever.WorldServer/Network/Message/Model/ClientHousingFlagsUpdate.cs
@@ -1,0 +1,23 @@
+﻿using NexusForever.Shared.Network;
+using NexusForever.Shared.Network.Message;
+using NexusForever.WorldServer.Game.Housing.Static;
+using NexusForever.WorldServer.Network.Message.Model.Shared;
+using System;
+
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    [Message(GameMessageOpcode.ClientHousingFlagsUpdate)]
+    public class ClientHousingFlagsUpdate : IReadable
+    {
+        public TargetPlayerIdentity Identity { get; } = new TargetPlayerIdentity();
+        public ResidenceFlags Flags { get; private set; }
+        public uint Unknown0 { get; private set; }
+
+        public void Read(GamePacketReader reader)
+        {
+            Identity.Read(reader);
+            Flags = reader.ReadEnum<ResidenceFlags>(3u);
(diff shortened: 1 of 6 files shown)


Source/NexusForever.WorldServer/Network/Message/Model/ClientHousingFlagsUpdate.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class ClientHousingFlagsUpdate : IReadable
+    {
+        public TargetPlayerIdentity Identity { get; } = new TargetPlayerIdentity();
+        public ResidenceFlags Flags { get; private set; }
+        public uint Unknown0 { get; private set; }
+
+        public void Read(GamePacketReader reader)
+        {
+            Identity.Read(reader);
+            Flags = reader.ReadEnum<ResidenceFlags>(3u);
```
Rawaho: Where do the 3 bits come from?
Having a look at the packet writer in the client just shows the player identity and a uint32

Rawaho (changes requested):
Looks good, just a small query about the packet structure.

Source/NexusForever.WorldServer/Network/Message/Model/ClientHousingFlagsUpdate.cs
myyanf: So I can't say I have a good answer for where they come from as I'm just structuring this off of packet data and a guess, and not off of anything more useful. However, if I change this to just read the identity and read Flags with reader.ReadEnum<ResidenceFlags>(32u), I get the following message from the server (develop branch with only this PR added in):
[WARN][Session] Failed to read entire contents of packet ClientHousingFlagsUpdate
So I had assumed the three extra bits were the flag, and the uint32 was something else unknown at this time.
I'll be more than happy to switch this around so that the flags enum is read as 32bit, but unless I'm missing something obvious (and I well might be), I'd still need to read in the 3 bits to avoid the warning message. Any pointers on how to proceed would be appreciated.
Thanks much for your help and feedback.

kirmmin: @myyanf: https://gist.github.com/kirmmin/2eab3aeb4b98420cca5effe812677ca7 - This is the full packet handler from the client. I haven't dug into which value is which, but this should help you check.

Rawaho: I'm guessing the values are harvest and garden split related.
This packet is sent from SetNeighborHarvestSplit and SetNeighborGardenSplit as well.

PR #264, opened 2020-04-12
Social/Emotes: Fix invisible units, animations and targeting

kirmmin:
This fixes some of the bugs reported by community members around emotes. Namely: 
- no "targeting" support - where it would say "Soandso points at you." / "Soandso points at Thingy.".
- no animations played to other Players.
- certain emotes (usually ones without animation) would make the Player who used it invisible to other players until they did an emote that made them visible (usually one with an animation).

This also adds basic model animation support to spells. I have a feeling that this is probably not the full picture, but it works as a first step into having the entity animations on/during spells. Most of them already work, but there are some where the entity is crying or cheering, for example, that wasn't visible to other Players before now.


Rawaho:
I would say this is now mostly implemented in a53c2129171a8894ff81bd1a4ff58e5abcee61b2.
The main thing missing is the visual effects on spells, this needs additional work on the expedition branch.

PR #266, opened 2020-04-19
Entity/Player: Fix Teleport crash when server is loading target zone

kirmmin:


Rawaho:
A command really should check for the teleport before calling TeleportTo and show the error message to the player.
The TeleportTo should throw an exception.

kirmmin:
@Rawaho: Thanks. I was a little unsure the best way to handle this, but I can convert it to that pretty simply. Give me a few :)

kirmmin:
~Updated. Thanks for the quick review.~

Edit: Thought of another way of Teleporting. 1 moment.

kirmmin:
Updated. Again. Should be good now ;)

PR #269, opened 2020-04-22
Housing/Core: Remove Residence on Character Delete

kirmmin:


Source/NexusForever.WorldServer/Game/Housing/ResidenceManager.cs
```diff
@@ -144,7 +155,8 @@ public void RegisterResidenceVists(ulong residenceId, string owner, string name)
         /// </summary>
         public void DeregisterResidenceVists(ulong residenceId)
         {
-            visitableResidences.Remove(residenceId);
+            if (visitableResidences.ContainsKey(residenceId))
+                visitableResidences.Remove(residenceId);
```
Rawaho: You do not need to check for existence before removal, remove can be called safely even if the residence doesn't exist

Rawaho (changes requested):

Source/NexusForever.WorldServer/Game/Housing/ResidenceManager.cs
kirmmin: Removed the unnecessary check. Thanks for the pointer.

PR #271, opened 2020-04-22
Spells/Core: Basic Continuous Casting support

kirmmin:


kirmmin:
This is replaced by [spells-v2 branch](/NexusForever/NexusForever/tree/spells-v2).

PR #279, opened 2020-04-30
Prerequisite/Checks: Checks Update - April 2020

kirmmin:
- Includes TargetGroup expansion to include TargetGroupType 10 & 11 (which are TargetGroups targeting other TargetGroups) for the existing Creature TargetGroup Cache.
- Includes update for `QuestManager.MeetsPrerequisites` check. The preq_quest values are an OR check.

I went through a bunch of different "numerical range" checks that are listed after my research (https://hackmd.io/@kirmmin/r1DyU2uPI#Prerequisites) and confirmed that all the ones I checked actually inverted GreaterThan/LessThan. It wanted things to be Greater Than or Equal To when Comparison Id 4 was being used. I'm doing my best to make sure I add as much PreReq support as we go, and figured I'd adjust the Enum now and, if we run into issues later, we can investigate. The investigation I did involved over 50 quests through Northern Wilds and Levian Bay.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/AssetManager.cs
+++ b/Source/NexusForever.WorldServer/Game/AssetManager.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
 using System.Reflection;
 using NexusForever.Database.World.Model;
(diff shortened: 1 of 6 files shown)


Rawaho:
Closing as this has been rebased in #466

PR #283, opened 2020-05-17
Quests/Core: Checklist quests support - WIP

kirmmin:
Contributes towards #306 

For this to function properly, entities that are part of a Checklist Quest will need to have their database entries updated/re-dumped with the `QuestChecklistIdx` from a sniff included.

This is used in quests like the Housing Starting Quest (Exile: http://www.jabbithole.com/quests/housing-of-the-future-9160) where you have to interact with certain entities and they update the quest appropriately.

This also includes the Reputation Rewards being sent to the Player after completing a quest.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Quest/QuestObjective.cs
+++ b/Source/NexusForever.WorldServer/Game/Quest/QuestObjective.cs
@@ -1,15 +1,21 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using Microsoft.EntityFrameworkCore.ChangeTracking;
 using NexusForever.Database.Character;
 using NexusForever.Database.Character.Model;
 using NexusForever.Shared;
+using NexusForever.Shared.GameTable;
 using NexusForever.Shared.GameTable.Model;
 using NexusForever.WorldServer.Game.Quest.Static;
+using NLog;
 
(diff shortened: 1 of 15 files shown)


Source/NexusForever.WorldServer/Game/Quest/QuestObjective.cs
```diff
@@ -69,9 +80,55 @@ public QuestObjective(QuestInfo info, QuestObjectiveEntry entry, byte index)
...
+            if (IsChecklist() || UsesTargetGroups())
+                BuildTargets();
+
             saveMask = QuestObjectiveSaveMask.Create;
         }
 
+        /// <summary>
+        /// Builds the target ID list for this <see cref="QuestObjective"/>.
+        /// </summary>
+        private void BuildTargets()
```
Rawaho, line 92: Does this need to be built for every player objective? Could this not be cached?

kirmmin: Refactored the build to AssetManager and then the QuestObjective grabs its list if it should :)

kirmmin:
This PR needs updating with code from this commit: https://github.com/kirmmin/NexusForever/commit/d9b4c0aa8b4f8ee111d799aa2f4d646d9304f1d1

PR #284, opened 2020-05-17
Utility/Story: StoryBuilder utility to send lore and story data to the client

kirmmin:


Rawaho:
I'm wondering if rather than having a singleton manager that this should have a local player manager.
This way the current window that is being shown can be tracked by the server and script support can be added if required.
Do you think this will be necessary?

kirmmin:
Sounds fine to me, @Rawaho. It makes sense to do it that way. I will refactor :)

Issue #270, opened 2020-04-22
[Bug] Database Entities not loading when in SelectMany chain.


kirmmin:
**Date:** April 22nd, 2020

**Description:** As the title says, `CharacterDatabase.GetCharacters()` does not return entities when the entity is part of a chained `SelectMany()` statement. At least, without an `Include()`.

**Occurrence:** Always

**OS:** Windows 10

**Branch/Commit:** develop

**STR:**
1. Create a Costume, Add some Items, and set the Costume to your Equipped one.
2. Close the client.
3. Re-open the client, note that your equipped Costume will have save and in the DB, the Costume will exist, but the holo wardrobe will not show the costume. This is due to the Character.Costume entity being "empty" when loading.

1. Go to Algoroc, accept some quests, check your quest log for them to exist.
2. Close the client.
3. Re-open the client, note that the Quest Log will be empty. The quests (and any progress) exist in the DB, but the Character.Quest entity will be empty when loading.

**Observed:** None of the SelectMany chained entities make it through to the Player instance. This includes the Mail entities.

**Expected:** All entities load as expected and are visible in the client.

PR #286, opened 2020-05-22
Character creation and load changes

jehgo:
- Added correct starting locations for each race at character create
- Fixed the loading of Costumes, Quests and Mail entities; Issue #270


Rawaho:
Database fix I approve but the character locations shouldn't be hard coded like this.
I think as you have been made aware there is also #152 which contains an attempt at starting locations, could take some inspiration from there.

If you want to remove the character creation changes and just have the database changes I'll approve, once the start location changes are more fleshed out it can be a new PR.

jehgo:
Removed the hardcoded locations.
Also added an additional fix for the max character level bug I mentioned on discord, essentially just move calculation below the loading of the stats for the character so it actually has a level value to check
Edit: I'll squash into single commits for future PRs per Kirmmin's request

jehgo:
Closing this as I'll be opening a new PR containing the fixes in the one

Issue #218, opened 2019-11-05
[Bug] Housing - Decor - Placing decor not already in crate causes large offset


Hallowisp:
**Date:** Ongoing since the housing plug update

**Description:** Placement of decor is not functioning correctly when placing decor that does not already exist in the user's crate. This issue was seemingly introduced after the housing plug update. I've been having trouble replicating the offset in a way I can make sense of, but here's a (probably incomplete) list of things I've noticed:

--Placing a decor item from the housing vendor that does not already exist in the crate will cause the decor to jump a large distance (~1/4 across the skyplot). 
--X and z coordinates are affected, but seemingly not y. 
--Character placement, direction, camera angle, and rotation of the decor item all seem to affect the offset.
--Placing items already in the crate is functioning normally.
--This issue is present whether or not there is a house placed in the central plug.
--I can't see any kind of pattern regarding the offset itself. 
--This issue is also present when using the 'clone' feature of Katia Builder Toolkit, a user addon, when the cloned item does not already exist in the crate.
--The issue is present when using KBT to clone items not in the crate and not natively from the housing vendor (i.e. any crafted decor item or world drop). 

**Occurrence:** Always
**OS:** Windows 7
**Branch/Commit:** Kirmmin's test server

**STR:** 

1. Teleport to your skyplot
2. Enter decoration mode
3. Open the housing vendor
4. Place an item from the vendor that does not already exist in your crate
5. Note the offset that occurs when decor item is placed
6. Repeat as necessary with different items, locations, and camera angles

**Observed:** Placing decor not already in crate causes a large offset
**Expected:** All decor will be placed at the coordinates the user specifies, whether or not the decor item already exists in the crate

Hallowisp:
Here's a download link for the Katia Builder Toolkit addon, if needed: https://drive.google.com/file/d/1fDHnQe_YPZ-YHYg1-zafMX3DvfyCa17X/view?usp=sharing

myyanf:
I can't say it will work on the test server, but I've submitted a fix (PR #296) for the location problem in the develop branch. In the unpatched develop branch, when placing directly from the vendor, the item drops down a bit until you do a server restart. This is corrected by tracking the PlotIndex of the placed items.

PR #296, opened 2020-07-04
Housing PlotIndex support - addresses #218 in the develop branch

myyanf:
As mentioned in the commit message, the known issues are:
1) Currently requires the plotIndex database default to be set
to 0 in order to properly set the plotIndex when placing on PlotIndex=0
from the vendor.
2) When placing in PlotIndex=0, the Y value in the database is not what you
see in game. The position as placed in game does persist as expected across
server loads so long as the plotIndex in the database is properly set to 0
(see 1). If the plotIndex is set to Int.MaxValue instead of 0, the item will
raise from its intended location after a server restart.

### DIFF
--- a/Source/NexusForever.Game.Abstract/Housing/IDecor.cs
+++ b/Source/NexusForever.Game.Abstract/Housing/IDecor.cs
@@ -26,7 +26,7 @@ namespace NexusForever.Game.Abstract.Housing
         /// <summary>
         /// Move <see cref="IDecor"/> to supplied position.
         /// </summary>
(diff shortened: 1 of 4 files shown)


kirmmin (approved):
Looks good to me. 

@Rawaho, I would say we need to look into this worldPosition thing. I noticed this when working on decor entities. The Z and X and flipped locally, so calculating world positions from a local decor position requires you to take the localZ and use that in the worldX calculation. It's kinda crazy. It might help to have some sort of converter function rather than hacking code to flip things around (like the `IsValidPlotForPosition` method does, currently).

PR #300, opened 2020-09-20
Enabled !character level 50

alex-anderson156:
This previously didnt work as 50 is>= 50

You can now set a character directly to level  50.


kirmmin:
Good catch! :)

Please clean up this PR and remove all the redundant merges. I would suggest resetting the brand to the head of develop and then either cherry-picking the code or copy-pasting your change to a new file. Thank you!

PR #312, opened 2020-09-27
Spells/Core: Support cancel-only effects

kirmmin:
This adds support to the server to allow for permanent duration spells & effects, that are only removed by being forced to end - either by another spell / effect, or a user interaction (like clicking off a buff). This allows us to end spells from continuously running in the server.


kirmmin:
This is replaced by [spells-v2 branch](/NexusForever/NexusForever/tree/spells-v2).

PR #313, opened 2020-09-28
Prerequisites/Checks: Fix Achievement check

kirmmin:


xNxExOx (approved):
Only checks if `bool` that is received as `uint` is 0 or 1, otherwise it logs.

Rawaho:
From what I can see this scenario only effects single prerequisite 2239, which is checking if achievement 367 is equal to 3.
This achievement doesn't exist so this check seems moot, going to close, happy to reopen if more evidence is provided to say otherwise.

PR #315, opened 2020-10-01
Player/Movement: Fix Position/Rotation emit errors

kirmmin:
This resolves the issue where Players don't show up for other Players when the other player enters range until the first player moves. Complicated way to say: Now, when a player enters range of another, you will immediately see them!

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
+++ b/Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
@@ -109,7 +109,7 @@ namespace NexusForever.WorldServer.Game.Entity.Movement
             AddCommand(new SetPositionCommand
             {
                 Position = new Position(position)
-            }, true);
+            }, !(owner is Player));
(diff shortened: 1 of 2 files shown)


Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
```diff
@@ -109,7 +109,7 @@ public void SetPosition(Vector3 position)
             AddCommand(new SetPositionCommand
             {
                 Position = new Position(position)
-            }, true);
+            }, !(owner is Player));
```
Rawaho, line 112: What happens when this is emitted? Does it just double up the packets? Since I can see in HandleEntityCommand we send ServerEntityCommand from the handler.
The issue to me seems more about the fact we aren't create/updating the command and when we build the ServerEntityCreate it doesn't contain the correct command.
This to me all seems like a design failure in the way we send commands to the client and I'm not sure if this is how we should fix it.

PR #316, opened 2020-10-02
Entity/Movement: Fix Rotation on spline

kirmmin:


Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
```diff
@@ -118,15 +118,48 @@ public void SetPosition(Vector3 position)
...
             }, true);
         }
 
+        /// <summary>
+        /// Get the unit id from <see cref="SetRotationFaceUnitCommand"/>.
+        /// </summary>
+        public uint? GetFaceUnit()
+        {
+            SetRotationFaceUnitCommand command = GetCommand<SetRotationFaceUnitCommand>();
+            return command?.UnitId > 0 ? command?.UnitId : null;
```
Rawaho, line 160: Is there a need to check for 0 here since you are using nullable?

Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
```diff
@@ -163,6 +196,12 @@ public void LaunchSpline(ushort splineId, SplineMode mode, float speed)
             StopSpline();
             splinePath = new SplinePath(splineId, mode, speed);
 
+            // TODO: This forces the entity to face the direction of travel. Need a way to handle walking backwards.
+            AddCommand(new SetRotationDefaultsCommand
+            {
+                Blend = true
```
Rawaho, line 202: What does blend actually do at the client?

Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
```diff
@@ -248,6 +303,7 @@ public void StopSpline()
             // TODO: calculate spline gradient to set rotation on end
 
             commands.Remove(splineCommand);
+            commands.Remove(EntityCommand.SetRotationDefaults);
```
Rawaho, line 306: You are adding SetRotationDefaultsCommand above but removing it here, does this serve any purpose?

Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
```diff
@@ -240,6 +285,16 @@ public void StopSpline()
...
             });
 
+            AddCommand(new SetRotationDefaultsCommand
+            {
+                Blend = false
+            });
+
+            AddCommand(new SetRotationCommand
+            {
+                Position = new Position(splinePath.GetPreviousPosition().GetRotationTo(position))
```
Rawaho, line 295: This is suppose to be the current facing direction? Seems like a very round about way of doing it

Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
```diff
@@ -109,7 +109,7 @@ public void SetPosition(Vector3 position)
             AddCommand(new SetPositionCommand
             {
                 Position = new Position(position)
-            }, true);
+            }, !(owner is Player));
```
Rawaho, line 112: What is the reason for this?

Source/NexusForever.WorldServer/Game/Entity/WorldEntity.cs
```diff
@@ -19,7 +19,18 @@ public abstract class WorldEntity : GridEntity
...
+        public Vector3 Rotation
+        {
+            get => (this is Player || MovementManager == null) ? rotation : MovementManager.GetRotation();
+            set
+            {
+                if (this is Player || MovementManager == null)
+                    rotation = value; // TODO: Emit rotations nearby if Player and no client command emitted?
+                
+                MovementManager?.SetRotation(value);
+            }
```
Rawaho, line 31: Not sure if we actually want setting the rotation to trigger the movement manager.
For example before a teleport, the rotation is set, this would trigger an entity command.

Rawaho (changes requested):
My suggestion to handling this would be tracking the angle to the next spline point in both directions, this can be calculated when the spline is created (in fact the code is already there for it)
This would also actually allow for proper handling with keys rather than default which retail didn't do

Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
kirmmin: It makes the entity face the direction it's walking in. My packet logs showed this when entities were in combat, or walking random directions - I didn't have the keys stuff happening.

Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
kirmmin: Correct. What would be your suggestion?

Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
kirmmin: Nope. This was me testing, and it's not necessary to remove on here.

Source/NexusForever.WorldServer/Game/Entity/Movement/MovementManager.cs
kirmmin: This is so we set the command for the Player but do not emit it. This only matters for the EntityCreate packet when another Player runs into range. See #315

Rawaho:
Resolved in 68d03abac5d86a496cb14a30faedc4ddec2c39cc

Issue #227, opened 2019-11-14
[Bug] Character Deletion does not delete Residence


kirmmin:
Date: November 14th, 2019

Description: When a character is deleted, the Residence doesn't get removed from the DB, nor does the ResidenceMap instance get closed and removed from game cache.

Occurrence: All.

OS: Windows 10

Branch/Commit: develop

STR:

1. Boot up a WorldServer using the develop branch
2. In character select, confirm a character has a residence and then delete it.
3. Note that, in the Database, that residence has not been removed.
3a. If the residence instance had been created, it will still be stored in the residence map cache.
3b. The residence of that character will also remain in the nearby houses list accessible from another player's plot. However, trying to port there will end with a forever-loading screen.

Observed: The character is "deleted" but the Residence remains in place.

Expected: The character's residence is deleted when the character is deleted.

PR #317, opened 2020-10-02
Character/Housing: Implement Residence deletion on character delete

MaxtorCoder:
This fixes issue #227 where Residences were not deleted from the database once the character got deleted.


Rawaho:
This will be handled in the up coming communities feature.
Residences will not be removed from the database and instead will just be marked as deleted.

PR #318, opened 2020-10-02
Item: Fix a special case where BagIndex can be out of sync

MaxtorCoder:
- Fixed the special case where `item.BagIndex` can be out of sync of the assigned Bag slot.
- Wrapped character loading into a try catch statement so the character enum can continue loading while there is an error

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Entity/Bag.cs
+++ b/Source/NexusForever.WorldServer/Game/Entity/Bag.cs
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using NexusForever.Database.Character;
 using NexusForever.WorldServer.Game.Entity.Static;
 using NLog;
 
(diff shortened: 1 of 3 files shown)


Source/NexusForever.WorldServer/Game/Entity/Bag.cs
```diff
@@ -23,6 +24,22 @@ public Bag(InventoryLocation location, uint capacity)
...
+        public void Save(CharacterContext context)
+        {
+            for (uint i = 0; i < items.Length; ++i)
+            {
+                // Item does not exist in slot
+                if (items[i] == null)
+                    continue;
+
+                // Item has incorrect slot, setting the BagIndex to the current index
+                if (i != items[i].BagIndex)
```
Rawaho, line 36: Maybe log this as well?
This is just hiding the problem so maybe log some more detailed information.

Source/NexusForever.WorldServer/Network/Message/Handler/CharacterHandler.cs
```diff
@@ -31,11 +31,14 @@
 using Item = NexusForever.WorldServer.Game.Entity.Item;
 using Residence = NexusForever.WorldServer.Game.Housing.Residence;
 using NetworkMessage = NexusForever.Shared.Network.Message.Model.Shared.Message;
+using NLog;
```
Rawaho: Ordering

Issue #307, opened 2020-09-24
Entity System - Properties, Stats, States


kirmmin:
This issue focuses on the features needed to support Properties and Stats for entities. This will not support all of them, just the basic, most appropriate ones needed for Milestone 1.

#### Items

- [x] Support Stat adjustment
   - [ ] Health
   - [ ] Focus
   - [ ] Class Resource
   - [x] Dash
- [x] Support Property adjustment
   - [ ] Max Health
   - [ ] Attack & Support Ratings
   - [ ] Armor
   - [ ] Focus Pool
   - [ ] Class Resource Caps
   - [x] Resource7 (Dash Pool)
   - [ ] Movement Speed
- [x] States
   - [x] Combat State
- [ ] Support adjustments from Items
   - [ ] Adding/Removing Items should adjust appropriately
- [ ] Support adjustments from spell effects
   - [ ] Apply adjustment on spell effect application
   - [ ] Remove appropriate adjustment on spell effect removal

#### Notes
- Support adjustments from items **does not** mean we are adding support for adding runes, set bonuses, or all properties/stats (like crit rating or any others).
   - It also doesn't mean that we are going to be 100% accurate with properties/stats given by items. _Dev Note: This needs a lot of reversing, and likely a Rawaho or someone as skilled task. I've (Kirmmin) has spent countless hours investigating this and, while I have made some headway in the calculations, it is not 100% accurate nor proven from reversing._
- The bare minimum of this would be Health Adjustment based on items and spells (both healing and damage).

kirmmin:
Properties were implemented as part of #179, but it's only a first stab and needs review.

Issue #302, opened 2020-09-24
Entity System - Threat Management, Leashing


kirmmin:
This issue captures the need for Threat Management for Entities. The purpose is to capture all items necessary to support Milestone 1.

#### Items

- [x] Threat list managed per entity
   - [x] Ability to add and remove entities from threat list
   - [x] Manage specific entity's threat level
   - [x] Clear threat list
- [ ] Able to grab all, specific (list position, threat value, or GUID/CharacterId), or random entities from threat list
- [x] Emit events on threat position changes
- [x] Set target and send packet(s) appropriately
- [x] Threat Commands
   - [x] Get list of entities and threat values
   - [x] Add entity to threat list
   - [x] Remove entity from threat list
   - [x] Adjust entity's threat value
   - [x] Adjust entity's list position

#### Notes
- This is purposefully a basic threat system. For M#1, we are just building the basic components to get things off the ground. Adjusting later or refactoring may or may not be necessary, but the plan is to keep it simple, so that if there is a need for complex support later, refactoring will be less painful.

kirmmin:
## Known Packet Structs

### `0x0908` Server0908
```
class Server0908
{
    public uint UnitId;
    public uint NewTargetId;
    public uint ThreatLevel;
}
```

#### Notes
- Seems to be sent at the start of the fight when the server-controlled unit (`UnitId`) enters combat. 
   - `ThreatLevel` is always `1` at the start of a fight for the Player who is capturing the log.
- Unknown how client responds.

### `0x0909` ServerThreatListUpdate
```
class ServerThreatListUpdate 
{
    uint SrcUnitId;
    uint[] ThreatUnitIds = new uint[5];
    uint[] ThreatLevels = new uint[5];
}
```

#### Notes
- This seems to be used throughout a fight to update Player's threat levels for this unit.
- It sends a fixed-size array of 5 units and their levels - there is no count field.
- Assumed that multiple of these packets are sent in raids and other things.
- Unknown how client responds.

### `0x090A` ServerThreatLevelUpdate
```
class ServerThreatLevelUpdate 
{
    uint UnitId;
    uint TargetId;
    uint ThreatLevels;
}
```

#### Notes
- This seems to be used mainly at the end of a fight to remove the Player (`TargetId`) from the server-controlled entity's (`UnitId`) threat list. 
- Unknown how client responds.

PR #320, opened 2020-10-04
Combat/Threat: Threat Manager implementation

kirmmin:
Implementation of a Threat Manager to be used as part of Milestone 1.

Contributes towards #302, #307 

- Includes target selection (and re-selection).
- Includes basic combat tracking, allowing for entities to be considered in or out of combat.
- Emits all packets as necessary for threat tracking.
- **Commands**
   - All commands are available as subcommands to `!entity threat`.
   - `!entity threat add <amount>` - Adds the supplied amount of threat to the target. Engaging them with you in combat.
      - Adding other players and adding more threat influences the Entity to change target.
   - `!entity threat remove <unitId>` - Removes the supplied unit id from the target's threat list completely. If you leave unit id blank, it will remove you instead.
   - `!entity threat clear` - Clears the target's threat list.
   - `!entity threat list` - Prints out the target's threat list, including all entities and their current threat amounts.


Source/NexusForever.WorldServer/Game/Combat/HostileEntity.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Returns the <see cref="WorldEntity"/> that this <see cref="HostileEntity"/> is associated with.
+        /// </summary>
+        public WorldEntity GetEntity(WorldEntity requester)
+        {
+            if (requester == null || requester.Map == null)
+                return null;
+
+            return requester.Map.GetEntity<WorldEntity>(HatedTargetUnitId);
```
Therzok:
```suggestion
            return requester?.Map?.GetEntity<WorldEntity>(HatedTargetUnitId);
```

Source/NexusForever.WorldServer/Game/Combat/ThreatManager.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            DoAddThreat(hostile, threat);
+        }
+
+        /// <summary>
+        /// Returns a <see cref="HostileEntity"/> for the given <see cref="WorldEntity"/>.
+        /// </summary>
+        private HostileEntity GetHostile(WorldEntity target)
+        {
+            if (hostiles.TryGetValue(target.Guid, out HostileEntity hostileEntity))
```
Therzok: This is not thread safe, so it may cause runtime crashes if there are multiple threads working with the collection at the same time.

Source/NexusForever.WorldServer/Game/Entity/Player.cs
```diff
@@ -621,6 +621,16 @@ public override void AddVisible(GridEntity entity)
...
+
+            if ((entity is WorldEntity we)) {
+
+                if (we.InCombat)
+                    Session.EnqueueMessageEncrypted(new ServerUnitEnteredCombat
+                    {
+                        UnitId = we.Guid,
+                        InCombat = we.InCombat
+                    });
+            }
```
Therzok: A bit easier and nicer to read:

```suggestion
            if (entity is WorldEntity we && we.InCombat)
            {
                Session.EnqueueMessageEncrypted(new ServerUnitEnteredCombat
                {
                    UnitId = we.Guid,
                    InCombat = we.InCombat
                });
            }
```

Source/NexusForever.WorldServer/Game/Entity/WorldEntity.cs
```diff
@@ -393,5 +403,59 @@ public virtual Disposition GetDispositionTo(Faction factionId, bool primary = tr
...
+
+            SelectTarget();
+        }
+
+        /// <summary>
+        /// Invoked when this <see cref="WorldEntity"/> is asked to select a target for an attack.
+        /// </summary>
+        public void SelectTarget(IEnumerable<HostileEntity> hostiles = null)
+        {
+            if (this is Player)
```
Therzok: This could be abstract and implemented in all the WorldEntity subclasses or virtual and overriden where needed.

In example, `MailBox` should not have a threat manager, it seems like something specific to UnitEntity?

Source/NexusForever.WorldServer/Game/Combat/HostileEntity.cs
```diff
@@ -0,0 +1,45 @@
+﻿using NexusForever.WorldServer.Game.Entity;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NexusForever.WorldServer.Game.Combat
+{
+    public class HostileEntity
+    {
+        public WorldEntity HaterOwner { get; private set; }
```
MaxtorCoder: I would honestly rename this to `Owner`, `HaterOwner` sounds a bit weird lol

Source/NexusForever.WorldServer/Game/Combat/HostileEntity.cs
MaxtorCoder: Please no, I rather have multiple if checks than having `?.` everywhere.

MaxtorCoder (commented):
Just some small lil' nitpicks

Source/NexusForever.WorldServer/Game/Combat/HostileEntity.cs
Fabi: Don't listen to Max. He doesn't know what's good

Source/NexusForever.WorldServer/Game/Combat/ThreatManager.cs
kirmmin: Fixed with Concurrent Dict; I'm not sure if we'd ever have 2 threads in there, but better to be safe than sorry

Source/NexusForever.WorldServer/Game/Entity/WorldEntity.cs
kirmmin: I appreciate the review on all the pieces, but especially on this part. It was good insight in to how I should've abstracted further than I had. Thank you.

Source/NexusForever.WorldServer/Game/Combat/ThreatManager.cs
Rawaho: Not sure if I agree with this.
All units on a map are updated sequentially so as long as an entity only has targets on a single map (which it always should) there should be no threading issues

Rawaho:
I haven't looked over the PR in depth yet but a few comments so far.
- The "combat entity" is suppose to UnitEntity, is there a case where combat would be applicable to entities not inheriting this?
If this is the case then do you raise the idea of removing UnitEntity all together.
- Not sure I like the command handler EntityThreatCommandCategory inheriting from another, maybe have a builder class that can generate strings for entities and can be used else where rather than being in a handler?

Over all it looks good but I'll need to have a proper sit down with the PR at some point.

kirmmin:
> * The "combat entity" is suppose to UnitEntity, is there a case where combat would be applicable to entities not inheriting this?
>   If this is the case then do you raise the idea of removing UnitEntity all together.

I wasn't aware of this distinction. I guess I questioned why `UnitEntity` was there because (and without really looking properly) I just assumed that all entity types inherited from `UnitEntity`. If this is the case, I can adjust a lot of these things to account for it. I wasn't making the case for `UnitEntity` to be removed, I think I would've leaned towards _wanting_ a "combat entity", and there was one all along 👍 

> * Not sure I like the command handler EntityThreatCommandCategory inheriting from another, maybe have a builder class that can generate strings for entities and can be used else where rather than being in a handler?

Sounds good to me. I hadn't thought of abstracting outside of the handler, but I can make some sort of Entity Utility Class that would expose that type of functionality.

Thanks for the review!

Source/NexusForever.WorldServer/Game/Combat/ThreatManager.cs
```diff
@@ -0,0 +1,170 @@
...
+                updateInterval.Reset();
+            }
+        }
+
+        /// <summary>
+        /// Add threat for the provided <see cref="WorldEntity"/>.
+        /// </summary>
+        public void AddThreat(WorldEntity target, int threat)
+        {
+            // TODO: Add way to pass in the Spell that was the source
```
alex-anderson156: I am not sure I would pass in the Spell that was the source.

But leave the Threat Manager relatively dumb. i.e. its told how much to add / subtract.

Its not just the spell you need to take into account, but also the buffs on the invoker and the target.

I personally would have another class responsible for calculating the threat and just pass it into here.

kirmmin:
Updated following your fast review, @Rawaho. Good for you to review properly, time permitting.

PR #323, opened 2020-10-07
[AuthServer] Send first online server to client instead of first in the db

SilentCLD:
Title. Also the server will now send `NpLoginResult.NoRealmsAvailableAtThisTime` when no servers are online.

### DIFF
--- a/Source/NexusForever.AuthServer/Network/Message/Handler/AuthenticationHandler.cs
+++ b/Source/NexusForever.AuthServer/Network/Message/Handler/AuthenticationHandler.cs
@@ -44,7 +44,7 @@ namespace NexusForever.AuthServer.Network.Message.Handler
                 }
 
                 // TODO: might want to make this smarter in the future, eg: select a server the user has characters on
-                ServerInfo server = ServerManager.Instance.Servers.FirstOrDefault();
+                ServerInfo server = ServerManager.Instance.Servers.FirstOrDefault(s => s.IsOnline);
                 if (server == null)
                 {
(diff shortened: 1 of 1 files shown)


MaxtorCoder (commented):
You should honestly send all available servers and do checking on the server you select.

SilentCLD:
@MaxtorCoder I was only able to find one video from the original launch and it put them straight into a realm on a fresh account (https://youtu.be/s5wr7qjm71s?t=1753). If it is possible / retail to go to the realmlist off the bat let me know and I'll take a look at it (I never played so no idea tbh 👀).

Rawaho:
This change is fine since you can’t choose your realm at first

Source/NexusForever.AuthServer/Network/Message/Handler/AuthenticationHandler.cs
```diff
@@ -44,7 +44,7 @@ void SendServerAuthDenied(NpLoginResult result)
                 }
 
                 // TODO: might want to make this smarter in the future, eg: select a server the user has characters on
-                ServerInfo server = ServerManager.Instance.Servers.FirstOrDefault();
+                ServerInfo server = ServerManager.Instance.Servers.Where(s => s.IsOnline).FirstOrDefault();
```
Rawaho: This change is good but could be shortened further by placing the predicate in the FIrstOrDefault call.
```suggestion
                ServerInfo server = ServerManager.Instance.Servers.FirstOrDefault(s => s.IsOnline);
```

SilentCLD: Good catch, modified

Issue #309, opened 2020-09-24
Combat System - Damage Calculations, Combat Logs


kirmmin:
This issue focuses on the "numbers" where it comes to combat and appropriate feedback to the client. The purpose is to capture all the items necessary to support Milestone 1.

#### Items

- [x] Formula Investigation
   - [x] Attack Power/Rating -> Ability Damage
   - [x] Support Power/Rating -> Ability Damage
   - [x] Attack/Support Power/Rating vs. Armor
   - [x] Randomisation of damage (was this a thing?)
   - [x] Base ratings (if there are any?)
- [x] Provide Combat Logs and a simple system to allow for combat log building (akin to TC?)

#### Notes

 -

PR #324, opened 2020-10-09
Combat/Core: Damage Calculator & Combat Log support

kirmmin:
Completes #309 

I was unsure how best way to implement this in a way that was reviewable, so I settled for a separate Manager, essentially, that would be able to do the calculations and spit out the damage (as a first stab). I recognise that this does not cover every potential damage or combat calculation, but this gives a base to work with.

Some notes:
- A lot of these calculations are assumed. I haven't dug deep, so I relied on what the stat tooltips said and what information I could find on the Internet.
- I added some randomisation to the damage, but I'm unsure how this is done in other emulators. There may be a better way to randomise the damage within a range.

Definitely open to hearing feedback on the stat weights, diminishing returns, damage ranges, etc.


Source/NexusForever.WorldServer/Game/Spell/Spell.cs
```diff
@@ -294,6 +298,12 @@ private void SendSpellGo()
             foreach (SpellTargetInfo targetInfo in targets
                 .Where(t => t.Effects.Count > 0))
             {
+                if (targetInfo.Effects.Where(x => x.DropEffect == false).Count() == 0)
```
Therzok: .Any() would be more efficient than count

Rawaho:
Going to close this and merge #457 as this has already been rebased to rework.

PR #327, opened 2020-10-13
Prerequisite/Core: Evaluate AND/OR Support

kirmmin:
Also includes evaluation logic for Spell Caster checks (it needed the AND/OR functionality, so bundling)

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Prerequisite/PrerequisiteManager.cs
+++ b/Source/NexusForever.WorldServer/Game/Prerequisite/PrerequisiteManager.cs
@@ -49,7 +49,21 @@ namespace NexusForever.WorldServer.Game.Prerequisite
             PrerequisiteEntry entry = GameTableManager.Instance.Prerequisite.GetEntry(prerequisiteId);
             if (entry == null)
                 throw new ArgumentException();
+            
+            switch ((EvaluationMode)entry.Flags)
+            {
+                case EvaluationMode.EvaluateAND:
+                    return MeetsEvaluateAnd(player, prerequisiteId, entry);
+                case EvaluationMode.EvaluateOR:
+                    return MeetsEvaluateOr(player, prerequisiteId, entry);
+                default:
+                    log.Trace($"Unhandled EvaluationMode {entry.Flags}");
+                    return false;
+            }
+        }
 
+        private bool MeetsEvaluateAnd(Player player, uint prerequisiteId, PrerequisiteEntry entry)
+        {
             for (int i = 0; i < entry.PrerequisiteTypeId.Length; i++)
             {
                 var type = (PrerequisiteType)entry.PrerequisiteTypeId[i];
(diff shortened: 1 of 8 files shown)


Source/NexusForever.WorldServer/Game/Prerequisite/PrerequisiteManager.cs
```diff
@@ -58,12 +69,39 @@ public bool Meets(Player player, uint prerequisiteId)
...
+                if (succeeds)
+                    break;
+
+                var type = (PrerequisiteType)entry.PrerequisiteTypeId[i];
+                if (type == PrerequisiteType.None)
+                    continue;
+
+                PrerequisiteComparison comparison = (PrerequisiteComparison)entry.PrerequisiteComparisonId[i];
+                if (Meets(player, type, comparison, entry.Value[i], entry.ObjectId[i]))
+                    succeeds = true;
```
Rawaho: Any reason you aren't returning true right away?

Source/NexusForever.WorldServer/Game/Prerequisite/PrerequisiteManager.cs
```diff
@@ -58,12 +69,39 @@ public bool Meets(Player player, uint prerequisiteId)
...
+            }
+
+            if (!succeeds)
```
Rawaho: If check can removed if returning right away above as this far down will only be executed if none of the pre requisites are met.

Source/NexusForever.WorldServer/Game/Spell/Spell.cs
```diff
@@ -112,11 +112,15 @@ private CastResult CheckCast()
             return CastResult.Ok;
         }
 
-        private bool CheckPrerequisites()
+        private CastResult CheckPrerequisites()
         {
-            // TODO
-            if (parameters.SpellInfo.CasterCastPrerequisites != null)
+            if (!(caster is Player player))
```
Rawaho, line 118: Is this going to always be the case? I assume there will be _some_ checking eventually for non players.

Source/NexusForever.WorldServer/Game/Prerequisite/Static/PrerequisiteEntryFlag.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NexusForever.WorldServer.Game.Prerequisite.Static
+{
+    [Flags]
+    public enum PrerequisiteEntryFlag
```
Rawaho: Are you sure these are actually flags? Seems more like a type since 0 is used as a valid value.

Source/NexusForever.WorldServer/Game/Prerequisite/Static/PrerequisiteComparison.cs
```diff
@@ -4,9 +4,9 @@ public enum PrerequisiteComparison
     {
         Equal    = 1,
         NotEqual = 2,
-        LessThan = 3,
-        LessThanOrEqual = 4,
-        GreaterThan = 5,
-        GreaterThanOrEqual = 6
+        GreaterThanOrEqual = 3,
```
Rawaho, line 7: Since these comparison changes are quite different are you sure this is the case for all pre requisites?

Source/NexusForever.WorldServer/Game/Prerequisite/PrerequisiteManager.cs
```diff
@@ -50,6 +50,17 @@ public bool Meets(Player player, uint prerequisiteId)
             if (entry == null)
                 throw new ArgumentException();
 
+            if ((PrerequisiteEntryFlag)entry.Flags == PrerequisiteEntryFlag.EvaluateAND)
```
Rawaho: You sure these are flags? If not make switch.

Rawaho (changes requested):

Source/NexusForever.WorldServer/Game/Prerequisite/PrerequisiteManager.cs
kirmmin: Uh, no reason 👀 lol

Source/NexusForever.WorldServer/Game/Spell/Spell.cs
kirmmin: This will not always be the case. I ran into a couple Spells yesterday that had PreReq evaluation required, but they were NPC abilities, I'll throw a TODO in here, but at some point we will need to extend / refactor PreReq System to allow UnitEntity (maybe even WorldEntity?) to be used in evaluation.

Source/NexusForever.WorldServer/Game/Prerequisite/Static/PrerequisiteEntryFlag.cs
kirmmin: Yeah. Valid. I probably overthought it. I looked for anything that was not 0 or 1 in the Prerequisites TBL and turned up nothing, so I'll remove the Flags and change the logic to switch

Source/NexusForever.WorldServer/Game/Prerequisite/Static/PrerequisiteComparison.cs
kirmmin: I keep running into this problem when working with Prerequisites. I am certain that this is the right operations for the comparison mode. What is often the case is the objectId is compared against the value, and that aligns with this enum. I can remove it, should you so wish, but I am fairly confident in the assumption I've made here.

PR #328, opened 2020-10-15
Inventory/auto send mats to satchel

mjmac87:
Added check against ItemUpdateReason which only allows adding to satchel if not moving items from satchel to bag. From this I check if the itemEntry is a material and attempt to add the amount being added to the satchel and returning the leftover to be added to the bag.


MaxtorCoder:
Wrong branch hahah, merge into dev

mjmac87:
Apologies! Quite new to this!

PR #329, opened 2020-10-15
Auto move materials to satchel then the bag if full

mjmac87:
Added check against ItemUpdateReason which only allows adding to satchel if not moving items from satchel to bag. From this I check if the itemEntry is a material and attempt to add the amount being added to the satchel and returning the leftover to be added to the bag.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
+++ b/Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
@@ -321,6 +321,12 @@ namespace NexusForever.WorldServer.Game.Entity
             if (itemEntry == null)
                 throw new ArgumentNullException();
 
+            // auto move materials into satchel but not when moving mats to bag from satchel
+            if (reason != ItemUpdateReason.ResourceConversion) {
+                if (GameTableManager.Instance.TradeskillMaterial.Entries.SingleOrDefault(i => i.Item2IdStatRevolution == itemEntry.Id) != null)
+                    count = player.SupplySatchelManager.AddAmount(new Item(characterId, itemEntry, count, charges), count);
+            }
+
             Bag bag = GetBag(InventoryLocation.Inventory);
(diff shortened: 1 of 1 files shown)


Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -321,6 +321,12 @@ public void ItemCreate(Item2Entry itemEntry, uint count, ItemUpdateReason reason
             if (itemEntry == null)
                 throw new ArgumentNullException();
 
+            // auto move materials into satchel but not when moving mats to bag from satchel
+            if (reason != ItemUpdateReason.ResourceConversion) {
```
MaxtorCoder:
```suggestion
            if (reason != ItemUpdateReason.ResourceConversion) 
            {
```

Source/NexusForever.WorldServer/Game/Entity/Inventory.cs
```diff
@@ -321,6 +321,13 @@ public void ItemCreate(Item2Entry itemEntry, uint count, ItemUpdateReason reason
             if (itemEntry == null)
                 throw new ArgumentNullException();
 
+            // auto move materials into satchel but not when moving mats to bag from satchel
+            if (reason != ItemUpdateReason.ResourceConversion) 
+            {
+                if (GameTableManager.Instance.TradeskillMaterial.Entries.SingleOrDefault(i => i.Item2IdStatRevolution == itemEntry.Id) != null)
+                    count = player.SupplySatchelManager.AddAmount(new Item(characterId, itemEntry, count, charges), count);
```
xNxExOx, line 328:
```suggestion
                    count = count - player.SupplySatchelManager.AddAmount(new Item(characterId, itemEntry, count, charges), count);
```
because `AddAmount` returns how much was added, not the remainder

xNxExOx (changes requested):

PR #330, opened 2020-10-23
Entity/AI: Range tracking other entities

kirmmin:


Source/NexusForever.WorldServer/Game/Entity/UnitEntity.cs
```diff
@@ -118,5 +119,41 @@ public void CancelSpellCast(uint castingId)
...
+            foreach (GridEntity entity in visibleEntities.Values)
+                CheckEntityRange(entity);
+        }
+
+        private void CheckEntityRange(GridEntity entity)
+        {
+            if (!(entity is WorldEntity we))
+                return;
+
+            if (!(this is Player))
```
Therzok: This should be an override in player.

Source/NexusForever.WorldServer/Game/Entity/WorldEntity.cs
```diff
@@ -393,5 +398,63 @@ public virtual Disposition GetDispositionTo(Faction factionId, bool primary = tr
...
+        {
+            return Position.GetDistance(target.Position) < RangeCheck;
+        }
+
+        /// <summary>
+        /// Returns whether the provided <see cref="WorldEntity"/> has entered this <see cref="WorldEntity"/>'s range.
+        /// </summary>
+        protected bool HasEnteredRange(WorldEntity target)
+        {
+            return inRangeEntities.Keys.Contains(target.Guid);
```
Therzok:
```suggestion
            return inRangeEntities.ContainsKey(target.Guid);
```

kirmmin: Could you elaborate on the difference here? I assume one is native and one is LINQ?

Therzok: Yeah, so, I double checked here, because I remembered something from older .NET versions.

`dict.ContainsKey` is going to do a O(1) lookup in the dictionary.

`dict.Keys.Contains` will do a linear aka O(n) lookup due to the fact that KeyCollection is not a hash-based collection, but just an array.

But it seems like `dict.Keys.Contains` is the same on .NET 5 (and older versions?) https://source.dot.net/#System.Private.CoreLib/Dictionary.cs,1424

Therzok: Also, one of my usual preferred reasons is: reduced cognitive load.

`Dictionary.ContainsKey` is a pretty standard API and well known behaviour.

By adding `.Keys`, there are increased considerations to take into account, like Keys implementation and behaviour.

kirmmin: If there is no performance benefit, I'll leave it up to @Rawaho to decide the best choice for this so that it stays consistent with rest of framework

MaxtorCoder: O(1) vs O(n) is a big difference tho, `ContainsKey` is the better way here.

PR #332, opened 2020-11-06
fix login without character

xNxExOx:
`GetAccountEntitlement` returns null for ne account without any character


Rawaho:
I have gone fixed this issue properly here aeec6009d8d0279dbaaf3c1efa926c1de05422f8
A proper fix needed the reward property changes which have also been merged.

PR #335, opened 2020-12-28
Added Endpoint to Trace output for Client connected/disconnected 

Digitalroot:
I found adding the endpoint to the trace info helped with debugging/trouble shooting.


Digitalroot:
Recreated as #337

PR #338, opened 2020-12-29
Windows Auto Restart Server Scripts

Digitalroot:
Added batch files to launch and auto restart the Auth, STS and World Server


Rawaho:
Thank you for the PR.
The end goal is to allow the server to run as a service and have the host decide how to handle a restart.
While the batch files are nice, they shouldn't be used in any proper hosting solution and it is unlikely they will benefit local or development use.

PR #339, opened 2020-12-29
Realm Shutdown command

Digitalroot:
Added Realm Shutdown command.

### Example Usage: 
`!realm shutdown 120`

This will cause the World Server to: 

1. Broadcast a message to warn all users that the server is shutting down. e.g. _"Realm is shutting down in 10 minutes."_
2. Start a count down timer. 
3. Rebroadcast a message, every 60 secs, to warn all users that the server is shutting down. e.g. _"Realm is shutting down in 9 minutes."_
4. When the countdown has less then 60 seconds. Rebroadcast time is reduced to 10 seconds. e.g. _"Realm is shutting down in 50 seconds."_
5. When the countdown ends. All connected sessions are disconnected and NexusForever.WorldServer.exe ends execution.


Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
```diff
@@ -1,23 +1,143 @@
...
+                timer.Start();
+            }
+        }
+
+        private static void OnShutdown(object source, ElapsedEventArgs e)
+        {
+            log.Info("Realm Shutdown.");
+            Timer timer = (Timer)source;
+            timer?.Stop();
+            Environment.Exit(0);
```
Rawaho: I wouldn't use `Environment.Exit`, the proper way to handle this would be shutting down each manager and cleanly exiting the main thread to prevent any data loss.

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
```diff
@@ -1,23 +1,143 @@
...
+            });
         }
+
+        private static void Shutdown()
+        {
+            log.Info("Realm is shutting down now.");
+            // Gracefully disconnect all users.
+            try
+            {
+                Parallel.ForEach(NetworkManager<WorldSession>.Instance.GetSessions(), session =>
```
Rawaho: This more than likely isn't thread safe and could cause issues.
For example this will eventually call `Player::CleanUp` which calls a few managers that aren't thread safe.
I would remove the Parallel here, a small delay in cleaning up each player during shutdown isn't a huge issue.

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
```diff
@@ -1,23 +1,143 @@
...
+                {
+                    session.Disconnect();
+                });
+            }
+            catch (InvalidOperationException exception)
+            {
+                if (!exception.Message.Equals("Collection was modified after the enumerator was instantiated."))
```
Rawaho: Any reason you are checking for this exception in particular? Could be related to the comment above?

Rawaho (changes requested):
A good PR overall, just a few small issues.

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
Digitalroot: Do the managers support shutting down?

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
Digitalroot: I will change it.

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
Rawaho: No, more than likely this will need to be added to the WorldManager which would propagate this to the other managers to do cleanup.

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
Digitalroot: If the session HashSet is altered while trying to iterate over it, the error gets thrown. This is unrelated to the ThreadSafe issue above.

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
Digitalroot: I can start a framework to handle this.

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
Fabi: You can't check for this string. It can change depending on the language

Digitalroot: I have changed this with c1eb63a

Source/NexusForever.WorldServer/Command/Handler/RealmCommandCategory.cs
Digitalroot: I added a shutdown Interface and added it to all the manager classes. A few of them already had a Shutdown method implemented. On shutdown the Shutdown method is called on each manager in the reverse order they are initialized. I'll leave the implementation of what each manager should do during shutdown to others more familiar with what each manager should do once shutdown. 

There is one issue with cleanly exiting the main thread. The console command has a `Console.ReadLine();` that is blocking. There is a way to workaround the block, but I do not think it will work on non-win platforms. I suggest using `Environment.Exit` in the `OnShutdown()` method I added to the WorldServer class. This method gets called after everything else is done. 

Thoughts?

Rawaho: I would move the ReadLine to a new thread, it calls HandleCommandDelay any way so thread safety isn't an issue

Digitalroot: I moved it to another thread (aa4f575) and it still blocks. Thread.Abort throws **PlatformNotSupportedException**

See https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread.abort?view=net-5.0

_PlatformNotSupportedException
.NET 5.0 and later and .NET Core: In all cases._

Digitalroot: It might be a good idea to remove the console from the WorldServer. It would prevent the WorldServer from running as a service.

Digitalroot: I was able to resolve the blocking issue by making the thread a background thread.

Rawaho:
This PR required a bit of work after the latest hosted service changes.
I have taken the shutdown command part of this PR and adapted it to the latest code base in 8518f523b541a966583d3873a5d013da1feda884.
Thank you for the PR 👍

PR #345, opened 2021-01-26
C#9 - Target Typing Improvements

MaxtorCoder:
https://medium.com/swlh/an-introduction-to-the-new-features-in-c-9-305dc8fb74d2

Change all `new <type>` to `new();`

**Example:**
```cs
// Before:
private List<int> numbers = new List<int>();

// Now:
private List<int> numbers = new();
```


Rawaho:
I like moving to the new C# 9 new typed syntax but I would prefer it only be for properties and fields, any local method variables should still use var with the right side typed to keep it consistent.

PR #348, opened 2021-03-28
Player/ZoneMap: Map complete support & distance bug fix

kirmmin:


kirmmin:
I changed the rate limiting check to use `Vector3.Distance` due to bugs I was experiencing during development. It would lock up in a position where all 3 were valid after you were moving in a certain direction for too long, and it would never update the `lastPosition` appropriately until you logged out and back in.

Rawaho:
Might pay to move most of this hex information into a template class at some point to prevent iterating the game tables too much as we additional stuff like this but that can be a job for another day.

PR #349, opened 2021-06-16
Fixed a network issue where, if the 4096-byte buffer filled up fully …

Dekker3D:
…and the 4 bytes starting a packet straddled the end of the buffer, it would cause disconnections. Often caused failed logins on characters on big plots.


Rawaho:
I have replaced this with af13b6ef884a259b5e918169babcfdc5f8e9a9ad
I didn't want to include the pipeline dependency just for this fix
Let me know if it doesn't do the trick!

PR #354, opened 2021-08-09
POC: Map Load Improvements - WIP

kirmmin:
**This is a proof of concept - not for merge!**

There's a number of issues with map performance at the moment, namely:
- When teleporting, the entire server waits for new map to be created. All data received does not get processed until after the new map has loaded.
- Players are stuck walking around in a map they won't be a part of while no processing occurs.
- - If, for whatever reason, the map fails to load, the server can experience bad state. They would've been removed from the existing Map on the server, and not added to the new one, but they would not be removed from game.
- Load screens are not being utilised correctly. We're not passing in all entities, setting up entities, spells, etc. properly before the player is put into the world, and that can sometimes cause the client to freak out or "be laggy" upon loading in.
   - Note; This is not resolved as part of this proof of concept, but would be a lot easier to accomplish with this type of framework.

This work focuses on allowing individual map instances to Initialise within their own update loop, and take advantage of "instant teleports" having the Player wait in the loading screen to be added to the new map, all the while taking strain off of the previous map. I experience 5-10s load times for most maps on my dev box and this greatly improved performance client side. We would need async map updates for this to work most effectively, which I know is coming.

This is my hacking as per our conversation yesterday, @Rawaho

### DIFF
--- a/Source/NexusForever.Shared/Network/GameSession.cs
+++ b/Source/NexusForever.Shared/Network/GameSession.cs
@@ -153,7 +153,7 @@ namespace NexusForever.Shared.Network
             IReadable message = MessageManager.Instance.GetMessage(packet.Opcode);
             if (message == null)
             {
-                log.Warn($"Received unknown packet {packet.Opcode:X}");
+                //log.Warn($"Received unknown packet {packet.Opcode:X}");
                 return;
             }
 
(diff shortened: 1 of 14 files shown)


Rawaho:
Going to close this.
The way maps and instances are handled really needs a complete relook, especially with instanced content coming soon:tm:.

PR #355, opened 2021-08-11
Social/Core: Fix Whisper echo

kirmmin:
This fixes a couple errors associated with whispers.

**Channel Switch**
When messaging someone then switching to write in /say (or any other channel), your echo message would be your previously sent whisper but to yourself.

**Message not sent response**
When a whisper is not sent, for whatever reason, it was being responded with manually. This has been corrected to use the correct packet.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Social/GlobalChatManager.cs
+++ b/Source/NexusForever.WorldServer/Game/Social/GlobalChatManager.cs
@@ -262,9 +262,19 @@ namespace NexusForever.WorldServer.Game.Social
 
         private void SendChatAccept(WorldSession session)
         {
+
             session.EnqueueMessageEncrypted(new ServerChatAccept
             {
                 Name = session.Player.Name,
+                Guid = session.Player.Guid
+            });
+        }
+
+        private void SendChatAccept(WorldSession session, string targetName)
+        {
+            session.EnqueueMessageEncrypted(new ServerChatAccept
+            {
+                Name = targetName,
                 Guid = session.Player.Guid
             });
         }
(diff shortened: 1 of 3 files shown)


Rawaho:
The boolean determines which channel it passes to the LUA function.
Either 6 or 34 depending on whether it is set, see the client code below:
```
channelType = ChatChannelTypeWhisper;
if ( packet->IsAccountWhisper )
  channelType = ChatChannelTypeAccountWhisper;
```

Rawaho:
The other value I'm not too sure on but it seems to be an array index for an array in the chat channel manager, after it does something with the value at that index it sets it to 0.
Would need further investigation :)

Source/NexusForever.WorldServer/Game/Social/GlobalChatManager.cs
```diff
@@ -340,40 +343,45 @@ ChatResult GetResult()
...
-            if (session.Player.Name == target.Name)
+            bool CanWhisper()
             {
-                SendMessage(session, "You cannot send a message to yourself.");
-                return;
+                if (target == null)
+                    return false;
+
+                if (session.Player.Name == target.Name)
+                    return false;
```
set4812, line 360: I lack of knowledge C# but I think this is simpler
```suggestion
                if (target == null || session.Player.Name == target.Name)
                    return false;
```

MaxtorCoder: Definitely

Source/NexusForever.WorldServer/Game/Social/GlobalChatManager.cs
```diff
@@ -260,12 +260,15 @@ public void HandleClientChat(WorldSession session, ClientChat chat)
             }
         }
 
-        private void SendChatAccept(WorldSession session)
+        private void SendChatAccept(WorldSession session, string nameOverride = null, uint? guidOverride = null)
```
Rawaho: Not really a fan of this, maybe have a function overload or only accept the name and guid directly for this rather than additional parameter?

Source/NexusForever.WorldServer/Network/Message/Model/ServerChatWhisperFail.cs
```diff
@@ -0,0 +1,20 @@
+﻿using NexusForever.Shared.Network;
+using NexusForever.Shared.Network.Message;
+
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    [Message(GameMessageOpcode.ServerChatWhisperFail)]
+    class ServerChatWhisperFail : IWritable
+    {
+        public string CharacterTo { get; set; }
+        public bool Unknown0 { get; set; }
```
Rawaho: IsAccountWhisper

Rawaho (commented):
Overall looks good, just a few comments

Source/NexusForever.WorldServer/Game/Social/GlobalChatManager.cs
kirmmin: It's intentionally 2 separate checks because It's cleaner to read individual checks and return early based on the check, and keeps in line with other portions of our codebase.

PR #357, opened 2021-08-15
Combat/Core: Take Damage & Death

kirmmin:
**Features**

- Entity Death (Players and Server-Controlled)
  - Entities DO NOT respawn until server restart
-  DeathState management and events
- Ghost creation & control (w/ logout and login support)
- Player Resurrection support (only "free" Wake Here's at the moment)

Notes: 
- **PLEASE READ:** This does not mean combat is now implemented! Things are definitely close, though. Soon™
- This was pretty tedious to "backward-apply" from my everything-branch to develop, so there are some pieces that will need to get re-added like Threat Adjustments #320 , Damage Calculations #309 , and such.


MaxtorCoder:
Also don't say things are close, because they're not. There are still way more systems that have to be implemented.

PR #326, opened 2020-10-11
Entity/Core: Vitals, Spell Costing, Dash implementation

kirmmin:
Contributes to #307,  #308 

Adds a few features on top of Vitals mapping:

- Vital `SpellEffect` support
- Innate costs (like Medic Cores, or Warrior's Kinetic Energy)
- Dash consumes energy, and there is a regeneration method that works 100% as Retail.
   - RegenMultiplier work by recovering X% (based on regen value) of the maximum every 0.5s.
- Unit & Player Events refactor

PR #358, opened 2021-08-18
Spells/Telegraphs: DamageFlag MVP & Esper Support

kirmmin:
**Features**

- Supports Telegraph DamageFlag which is a property that appears to be calculated server side.  This mostly fixes the duplication of damage that we're seeing, currently. Clearly, Wildstar had multiple Telegraphs that were being checked (many either invisible or visible only in certain states). I did investigate DisplayFlags and only figured out 1 of the bits so far, but it's honestly not important at this time because we only care about tracking whether or not the telegraph should apply "damage" (or healing or buffs) to the right targets.
    - I went through about 100 Telegraphs and compared the Flags based on the caster, class, expectations of the spell (from it's text and videos) and narrowed down a few of the bits in the mask. 
    - They are, obviously, up to in interpretation - they don't seem to be used in the client at all, or if they are, they are deep in the code.
- This also adds some support for Esper Psi Points. With the #326 PR, Vitals will allow for spell costing and this extends that to consume the rest of the Esper's Psi Points and also determine the right effects to cast based on the current Psi Points.
- This also introduces the CastMethod Enum from the Mega Spells PR because I needed to check a CastMethod as part of the DamageFlag.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Spell/Spell.cs
+++ b/Source/NexusForever.WorldServer/Game/Spell/Spell.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using NexusForever.Shared;
 using NexusForever.Shared.GameTable.Model;
 using NexusForever.WorldServer.Game.Entity;
+using NexusForever.WorldServer.Game.Entity.Static;
 using NexusForever.WorldServer.Game.Prerequisite;
 using NexusForever.WorldServer.Game.Spell.Event;
 using NexusForever.WorldServer.Game.Spell.Static;
@@ -20,6 +21,7 @@ namespace NexusForever.WorldServer.Game.Spell
         public uint CastingId { get; }
         public bool IsCasting => status == SpellStatus.Casting;
         public bool IsFinished => status == SpellStatus.Finished;
+        public CastMethod CastMethod { get; }
 
         private readonly UnitEntity caster;
         private readonly SpellParameters parameters;
@@ -36,6 +38,8 @@ namespace NexusForever.WorldServer.Game.Spell
             this.parameters = parameters;
             CastingId       = GlobalSpellManager.Instance.NextCastingId;
             status          = SpellStatus.Initiating;
+            CastMethod      = (CastMethod)parameters.SpellInfo.BaseInfo.Entry.CastMethod;
+
 
             if (parameters.RootSpellInfo == null)
                 parameters.RootSpellInfo = parameters.SpellInfo;
@@ -69,6 +73,10 @@ namespace NexusForever.WorldServer.Game.Spell
             CastResult result = CheckCast();
             if (result != CastResult.Ok)
             {
+                // TODO: Add back in with Continous Cast deployment
+                //if (caster is Player)
+                //    (caster as Player).SpellManager.SetAsContinuousCast(null);
+
                 SendSpellCastResult(result);
                 return;
             }
@@ -174,6 +182,7 @@ namespace NexusForever.WorldServer.Game.Spell
         private void InitialiseTelegraphs()
         {
             telegraphs.Clear();
+
             foreach (TelegraphDamageEntry telegraphDamageEntry in parameters.SpellInfo.Telegraphs)
                 telegraphs.Add(new Telegraph(telegraphDamageEntry, caster, caster.Position, caster.Rotation));
         }
@@ -194,6 +203,13 @@ namespace NexusForever.WorldServer.Game.Spell
                     CastResult     = result,
                     CancelCast     = true
                 });
+
+                if (result == CastResult.CasterMovement)
+                    player.SpellManager.SetGlobalSpellCooldown(0d);
+
+                //player?.SpellManager.SetAsContinuousCast(null);
+
+                SendSpellCastResult(result);
             }
 
             events.CancelEvents();
@@ -215,6 +231,10 @@ namespace NexusForever.WorldServer.Game.Spell
             ExecuteEffects();
             CostSpell();
 
+            // TODO: Add back in with Vitals
+            //if (caster is Player player && HasEsperCost())
+            //    caster.ModifyVital(Vital.Resource1, -caster.GetVitalValue(Vital.Resource1));
+
             SendSpellGo();
         }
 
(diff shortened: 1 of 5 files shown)


Source/NexusForever.WorldServer/Game/Spell/Telegraph.cs
```diff
@@ -67,12 +67,40 @@ private void ApplyRotationOffsets()
...
+                    targets.Remove(target);
+
             return targets.Select(t => t as UnitEntity);
         }
 
+        private bool EvaluateDamageFlagsForTarget(GridEntity target, Spell spell)
+        {
+            TelegraphDamageFlag damageFlag = (TelegraphDamageFlag)TelegraphDamage.TelegraphDamageFlags;
+
+            if ((damageFlag & TelegraphDamageFlag.SpellMustBeMultiPhase) != 0u)
```
MaxtorCoder:
```suggestion
            if ((damageFlag & TelegraphDamageFlag.SpellMustBeMultiPhase) != 0)
```

Source/NexusForever.WorldServer/Game/Spell/Static/CastMethod.cs
```diff
@@ -0,0 +1,21 @@
...
+        Channeled               = 0x0001,
+        PressHold               = 0x0002,
+        ChanneledField          = 0x0003,
+        UNUSED04                = 0x0004,
+        ClientSideInteraction   = 0x0005,
+        RapidTap                = 0x0006,
+        ChargeRelease           = 0x0007,
+        Multiphase              = 0x0008,
+        Transactional           = 0x0009,
+        Aura                    = 0x000A
```
MaxtorCoder, line 19: Is there a specific reason why you use hex for this?

Source/NexusForever.WorldServer/Game/Spell/Telegraph.cs
Fabi: .HasFlag is your friend

Source/NexusForever.WorldServer/Game/Spell/Static/TelegraphDamageFlag.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading.Tasks;
+
+namespace NexusForever.WorldServer.Game.Spell.Static
+{
+    /// <summary>
+    /// TelegraphDamageFlags in the TelegraphEntry TBL seem to all be calculated server side. Unable to find information in the client on them.
+    /// This is assumed value meaning based on complete speculation when comparing telegraphs.
+    /// </summary>
+    [Flags]
+    public enum TelegraphDamageFlag
```
Fabi, line 14: Either name it Flags at the end (which you should never do!) or just TelegraphDamage (there is no plural of damage)

Source/NexusForever.WorldServer/Game/Spell/Spell.cs
```diff
@@ -194,6 +203,13 @@ public void CancelCast(CastResult result)
                     CastResult     = result,
                     CancelCast     = true
                 });
+
+                if (result == CastResult.CasterMovement)
+                    player?.SpellManager.SetGlobalSpellCooldown(0d);
```
Fabi: I don't think that null check is required

Source/NexusForever.WorldServer/Game/Spell/Spell.cs
```diff
@@ -242,6 +262,15 @@ private void ExecuteEffects()
         {
             foreach (Spell4EffectsEntry spell4EffectsEntry in parameters.SpellInfo.Effects)
             {
+                // Check if Spell uses Psi Points, Confirm Effect is the right damage spell for the remaining Psi Points
+                if (HasEsperCost() && caster is Player esperPlayer && esperPlayer.Class == Class.Esper)
```
Fabi: I guess we have a double is Player & class check here. It's already done on HasEsperCost. GetStatFloat works on caster directly too I guess

Source/NexusForever.WorldServer/Game/Spell/Telegraph.cs
MaxtorCoder: ^ Honestly was not sure if that slwas some extension method or not :kekw:

Source/NexusForever.WorldServer/Game/Spell/Spell.cs
DanielMueller1996: thats correct, "caster is Player player" returns false if either is null.

Source/NexusForever.WorldServer/Game/Spell/Static/TelegraphDamageFlag.cs
kirmmin: TelegraphDamage is already a thing, and this is singular because it's a representation of a single flag. How it is currently named is in line with everything else in the codebase.

Source/NexusForever.WorldServer/Game/Spell/Static/CastMethod.cs
kirmmin: Easier to distinguish when comparing values to the reversed code.

MaxtorCoder: I mean, you can create enums in IDA to prevent this type of thing.

MaxtorCoder: Also it's a setting in IDA to show hex values, so honestly not a valid reason.

kirmmin: To be clear, my reasoning is not invalidated because you disagree with me. Either way, this has been common practice for the majority of the codebase. if we're taking enum values from code that's been reversed, we generally use hex codes. Look at SpellEffectTypes, CharacterFlag, or ItemProficiency.

I'm open to either way because it's just a representation of a value. I'll leave it up to @Rawaho to decide.

Fabi: Flags should be hex. everything else not unless it's about super big numbers like idk color codes for example. Or well.. big enums like packet stuff.

Source/NexusForever.WorldServer/Game/Spell/Telegraph.cs
MaxtorCoder: Also rename the enum to `TelegraphDamageFlags` plz thanks

Source/NexusForever.WorldServer/Game/Spell/Static/TelegraphDamageFlag.cs
Fabi: > TelegraphDamage is already a thing, and this is singular because it's a representation of a single flag. How it is currently named is in line with everything else in the codebase.

Well yea. The only reason to name it Flags is that it's named like that in the tbl file. However these flags are also combined and only non Flag enums should be singular :D

PR #373, opened 2021-08-25
Quests/Core: Improved logic to allow quest pickup

kirmmin:


Source/NexusForever.WorldServer/Game/Entity/QuestManager.cs
```diff
@@ -248,10 +248,25 @@ private bool MeetsPrerequisites(QuestInfo info)
             if (player.Level < info.Entry.PrerequisiteLevel)
                 return false;
 
+            bool preReqQuestsOrCheck = (info.Entry.PrerequisiteFlags & 1) != 0u;
+            bool preReqQuestsCompleted = info.Entry.PrerequisiteQuests.Where(q => q != 0).Count() == 0u;
```
Fabi: You can use .Any here

Another alternative would be to combine setting it to true/false only within the next foreach loop instead of doing a separate check and enumeration here.

Source/NexusForever.WorldServer/Game/Quest/QuestInfo.cs
```diff
@@ -46,7 +46,7 @@ public bool IsAutoComplete()
         /// </summary>
         public bool IsCommunicatorReceived()
         {
-            return (Entry.Flags & 0x08) != 0;
+            return (Entry.Flags & 0x08) != 0 || (Entry.Flags & 0x04) != 0;
```
Rawaho: IsCommunicatorReceived is a function in the client which only refers to flag 8.
What is flag 4 suppose to do?

kirmmin: This was a mis-copy. Let me fix.

Flag 4 appears to allow a Quest the "Call Back" functionality.

kirmmin: Removed now. This check is part of #374

Source/NexusForever.WorldServer/Game/Quest/Quest.cs
```diff
@@ -302,7 +302,7 @@ public void ObjectiveUpdate(uint id, uint progress)
             objective.ObjectiveUpdate(progress);
             SendQuestObjectiveUpdate(objective);
 
-            if (objectives.All(o => o.IsComplete()))
+            if (objectives.All(o => o.IsComplete()) && State != QuestState.Achieved)
```
Rawaho: Is there any reason this can't be at the start of ObjectiveUpdate?
I'm assuming objectives shouldn't be updated at all if the quest state is achieved?

kirmmin: No, no reason, just didn't occur to me :) I can make that change.

Rawaho: I was just thinking from both a performance and fail early point of view it makes more sense and would lead to less confusion :P

Source/NexusForever.WorldServer/Game/Entity/QuestManager.cs
Rawaho: Because I like LINQ I would have done something like:
``` csharp
IEnumerable<uint> quests = info.Entry.PrerequisiteQuests.Where(q => q != 0);
if ((info.Entry.PrerequisiteFlags & 1) != 0u)
	preReqQuestsCompleted = quests.Any(q => GetQuestState((ushort)q) == QuestState.Completed);
else
	preReqQuestsCompleted = quests.All(q => GetQuestState((ushort)q) == QuestState.Completed);
	
if (!preReqQuestsCompleted)
	return false;
```

Rawaho: Inb4 Fabian hates it 🤣

kirmmin: Using the LINQ for readability :) Thanks for the suggestions

Source/NexusForever.WorldServer/Game/Quest/Quest.cs
kirmmin: Updated.

Source/NexusForever.WorldServer/Game/Entity/QuestManager.cs
Fabi: > Inb4 Fabian hates it 🤣

Why is the quests field created then? you can query info.Entry.PrerequisiteQuests directly with  conditions and it's readable and nice too but you save a full enumeration.

PR #374, opened 2021-08-25
Quest/Core: Quest mentions & optional objective complete

kirmmin:
Solves problem when quest is mentioned, a user doesn't accept, and is unable to get the quest ever again (specifically, if they logout after not accepting).
Delivers the quest for "Call Back" from the client.
Allows tracking of all quests the user has been offered by the client.
Automatically completes Optional Objectives when the required step is complete.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Quest/Quest.cs
+++ b/Source/NexusForever.WorldServer/Game/Quest/Quest.cs
@@ -29,7 +29,7 @@ namespace NexusForever.WorldServer.Game.Quest
                 state = value;
                 saveMask |= QuestSaveMask.State;
 
-                OnStateChange(oldState, state);
+                OnStateChange(oldState);
             }
         }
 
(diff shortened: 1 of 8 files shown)


Source/NexusForever.WorldServer/Game/Quest/Quest.cs
```diff
@@ -320,8 +352,38 @@ private bool CanUpdateObjective(QuestObjective objective)
             return true;
         }
 
+        private bool AreAllRequiredObjectivesCompleted()
```
Rawaho: Is there a reason this is more complex then just checking if the objective is optional via flags?
Not quite sure I understand the current function.

Source/NexusForever.WorldServer/Game/Quest/Quest.cs
```diff
@@ -300,6 +328,10 @@ public void ObjectiveUpdate(uint id, uint progress)
                 return;
 
             objective.ObjectiveUpdate(progress);
+
+            if (AreAllRequiredObjectivesCompleted())
+                CompleteAllOptionalObjectives();
```
Rawaho: Was retail like this?
I can't remember if you could complete optional objectives after a quest was considered complete.

PR #375, opened 2021-09-04
Player/Create: Starting Locations, Boost Support

kirmmin:
- All locations were pulled from live logs
- Modifiable locations to support custom starting locations
- Boost is supported but tokens are disabled, still, due to needing bag support

Note: We need to store rotations against the character so it will spawn facing the right direction, but I will put that in a separate PR.

Thanks to @ahoiahoi for providing the original implementation of this that I referenced when updating this to latest version of NF.


kirmmin:
Marked this as WIP because the rotations aren't accurate now that I'm testing a rotations PR.

kirmmin:
Removed WIP, fixed up starting locations and rotations.

PR #384, opened 2021-11-23
[Inventory/Item] Fix IsEquippable incorrect return

kirmmin:
If SlotEntry was empty, I was getting "true" returns. This was most prevalent when creating a Veteran character, it would try to put the Protostar Rucksack lootbag in an equipped slot, even without a SlotEntry. 

Solution I chose was to update `IsEquppable` with a non-null check.

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Entity/ItemInfo.cs
+++ b/Source/NexusForever.WorldServer/Game/Entity/ItemInfo.cs
@@ -30,7 +30,7 @@ namespace NexusForever.WorldServer.Game.Entity
         /// </summary>
         public bool IsEquippable()
         {
-            return SlotEntry != null && SlotEntry?.EquippedSlotFlags != 0u;
+            return SlotEntry != null && SlotEntry.EquippedSlotFlags != 0u;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Source/NexusForever.WorldServer/Game/Entity/ItemInfo.cs
```diff
@@ -30,7 +30,7 @@ public ItemInfo(Item2Entry entry)
         /// </summary>
         public bool IsEquippable()
         {
-            return SlotEntry?.EquippedSlotFlags != 0u;
+            return SlotEntry != null && SlotEntry?.EquippedSlotFlags != 0u;
```
Therzok:
```suggestion
            return SlotEntry != null && SlotEntry.EquippedSlotFlags != 0u;
```

Therzok: Avoid 2nd null check :D

Fabi: You can also just use `return SlotEntry?.EquippedSlotFlags > 0u;`

PR #385, opened 2021-11-25
[Realm/Queue] Queue connections when at max capacity

kirmmin:
This is a basic Login Queue that handles admitting players. The average wait time is not calculated, just hardcoded as (position in queue * 30 seconds).

### DIFF
--- a/Source/NexusForever.WorldServer/Game/LoginQueueManager.cs
+++ b/Source/NexusForever.WorldServer/Game/LoginQueueManager.cs
@@ -1,29 +1,38 @@
-﻿using NexusForever.Shared;
+﻿using System;
+using System.Collections.Generic;
+using NexusForever.Shared;
 using NexusForever.Shared.Configuration;
 using NexusForever.Shared.Game;
 using NexusForever.Shared.Network;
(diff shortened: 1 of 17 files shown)


Source/NexusForever.WorldServer/Game/LoginQueueManager.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            return true;
+        }
+
+        private void SendQueueStatus(uint queuePosition, WorldSession session)
+        {
+            session.EnqueueMessageEncrypted(new ServerQueueStatus
+            {
+                QueuePosition = queuePosition,
+                WaitTimeSeconds = (uint)(queuePosition * 30f)
```
Matt-Hurd, line 194: I feel like even when hardcoded, this should take into account the total server size. A server with 10 capacity vs 1000 would have a different wait time for a person 10th in line. Something like `queuePosition / maximumPlayers * 50 * 30`. Although this is probably "good enough" for now, since an accurate estimation would need heuristics on average connection length at time of day, which is out of scope.

PR #387, opened 2021-11-30
[RBAC/Roles] Apply Default Roles

kirmmin:
`!account create` now includes an optional 3rd parameter of role for the user. If omitted, it will use the default role as specific in the WorldServer config, or failing that "1".

### DIFF
--- a/Source/NexusForever.WorldServer/Command/Handler/AccountCommandCategory.cs
+++ b/Source/NexusForever.WorldServer/Command/Handler/AccountCommandCategory.cs
@@ -1,4 +1,5 @@
-﻿using NexusForever.Shared.Cryptography;
+﻿using NexusForever.Shared.Configuration;
+using NexusForever.Shared.Cryptography;
 using NexusForever.Shared.Database;
 using NexusForever.WorldServer.Command.Context;
 using NexusForever.WorldServer.Command.Convert;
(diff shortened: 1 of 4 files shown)


AsterBunnymund:
!account create

Source/NexusForever.WorldServer/Command/Handler/AccountCommandCategory.cs
```diff
@@ -14,16 +15,21 @@ public class AccountCommandCategory : CommandCategory
...
+            [Parameter("Role")]
+            uint? role = null)
         {
             if (DatabaseManager.Instance.AuthDatabase.AccountExists(email))
             {
                 context.SendMessage("Account with that username already exists. Please try another.");
                 return;
             }
 
+            if (role == null)
```
Rawaho: Null coalescing assignment operator is your friend here :D

kirmmin: ??=  ? ;)

MaxtorCoder: yessir

kirmmin: Operator is now in use.

PR #388, opened 2021-12-02
[Database/Auth] Fix load time issues

kirmmin:
Was seeing Query timeouts on the Test Server, and believe it was caused by the structure of the previous implementation and all the `Include()` use, as we know that caused issues before.

This unlocked the accounts that were experiencing issues.


kirmmin:
Refactored to use `AsSpliQuery()`.

Issue #104, opened 2019-03-02
Holo Wardrobe InvalidOperationException


badbaki:
This doesn't crash the server or disconnect players but I've noticed it being thrown whenever someone edits a costume that isn't equipped. Users reported unlocks and costumes not being saved after encountering the error. They'd log out and back in and nothing would be saved. 

> [ERROR][Session] System.InvalidOperationException: Operation is not valid due to the current state of the object.
> at NexusForever.WorldServer.Game.Entity.CostumeManager.SetCostume(Int32 index) in C:\Users\Baki\Desktop\NF-Dev\NexusForever\Source\NexusForever.WorldServer\Game\Entity\CostumeManager.cs:line 190
> at NexusForever.WorldServer.Network.Message.Handler.CostumeHandler.ClientCostumeSet(WorldSession session, ClientCostumeSet costumeSet) in C:\Users\Baki\Desktop\NF-Dev\NexusForever\Source\NexusForever.WorldServer\Network\Message\Handler\CostumeHandler.cs:line 17
> at NexusForever.Shared.Network.GameSession.HandlePacket(ClientGamePacket packet) in C:\Users\Baki\Desktop\NF-Dev\NexusForever\Source\NexusForever.Shared\Network\GameSession.cs:line 186 System.InvalidOperationException: Operation is not valid due to the current state of the object.
> at NexusForever.WorldServer.Game.Entity.CostumeManager.SetCostume(Int32 index) in C:\Users\Baki\Desktop\NF-Dev\NexusForever\Source\NexusForever.WorldServer\Game\Entity\CostumeManager.cs:line 190
> at NexusForever.WorldServer.Network.Message.Handler.CostumeHandler.ClientCostumeSet(WorldSession session, ClientCostumeSet costumeSet) in C:\Users\Baki\Desktop\NF-Dev\NexusForever\Source\NexusForever.WorldServer\Network\Message\Handler\CostumeHandler.cs:line 17
> at NexusForever.Shared.Network.GameSession.HandlePacket(ClientGamePacket packet) in C:\Users\Baki\Desktop\NF-Dev\NexusForever\Source\NexusForever.Shared\Network\GameSession.cs:line 186

PR #389, opened 2021-12-02
[Entity/CostumeManager] Prevent double-setting costume index

Matt-Hurd:
Addresses: https://github.com/NexusForever/NexusForever/issues/104

The client calls `Save` and then `Set` in the case of editing a non-equipped costume. This caused a double `Set` which would error out and cause the client and server state to be desync'd.


kirmmin:
Does this resolve https://github.com/kirmmin/NexusForever/issues/27#issuecomment-983320262, too?

kirmmin:
> Does this resolve [kirmmin#27 (comment)](https://github.com/kirmmin/NexusForever/issues/27#issuecomment-983320262), too?

Yes, is the answer. 2 Birds, 1 stone! Good job :)

PR #391, opened 2021-12-02
[Guilds/Communities] Handle Residence(s) on Disband or Leave

kirmmin:
Updates the Community Visit list appropriately on Disband
Fix Players not having a Residence assigned to ResidenceManager when Residence is created
Update Communitiy Visit List on Map Enter in case Community was just disbanded

### DIFF
--- a/Source/NexusForever.WorldServer/Game/Map/ResidenceMapInstance.cs
+++ b/Source/NexusForever.WorldServer/Game/Map/ResidenceMapInstance.cs
@@ -89,6 +89,7 @@ namespace NexusForever.WorldServer.Game.Map
             if (entity is not Player player)
                 return;
 
+            GlobalResidenceManager.Instance.SendRandomVisitableCommunities(player.Session);
             SendResidences(player);
             SendResidencePlots(player);
             SendResidenceDecor(player);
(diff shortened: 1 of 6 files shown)


Source/NexusForever.WorldServer/Game/Map/ResidenceMapInstance.cs
```diff
@@ -89,6 +89,7 @@ protected override void AddEntity(GridEntity entity, Vector3 vector)
             if (entity is not Player player)
                 return;
 
+            GlobalResidenceManager.Instance.SendRandomVisitableCommunities(player.Session);
```
Rawaho, line 92: Any reason this is being sent here?
The client has a packet it sends to request this when the random property window is open.

kirmmin: Sorry, I hadn't seen this comment.

I wrote this in here because if a Community is Disbanded, anybody in the current Community Map is removed and then sent to their own plot. When they reload, if they click on the "View Visitable Communities" thing, the old Community shows up because it's using the values from the first time the window was loaded. I discovered that re-sending the visitable communities was all that was needed to remove the, now disbanded, community.

PR #393, opened 2021-12-03
[Server/Core] Calculate Server Time for client

kirmmin:


Rawaho:
Updated this to use TimeSpan offset rather than hours as some timezones are not hour aligned.

PR #395, opened 2021-12-03
[Entity/Vendor] Support Credit Cost formula

kirmmin:
The formula that I derived was determined by taking values as seen in game. I was unable to discover the formula in the client, but said I would push this as a PR as it gives a good baseline value to work with.
```
((((PowerLevel * PowerLevel) * ItemQualityId) * Type.VendorMultiplier) * Category.VendorMultiplier) * 1.125
```
The value always rounded down. It's worth mentioning that this value worked for items regardless of PowerLevel. Initially, I thought this would only work on Items between 20 and 50 PowerLevel, but finding the ItemCategory's VendorMultiplier was the missing piece that made this formula always work.

On top of this, Vendors sell items for `the above value / 0.14966`. I could not find any related GameTables to figure out how either multiplier was determined.

### DIFF
--- a/Source/NexusForever.Game.Abstract/Entity/IItemInfo.cs
+++ b/Source/NexusForever.Game.Abstract/Entity/IItemInfo.cs
@@ -26,6 +26,18 @@ namespace NexusForever.Game.Abstract.Entity
         ushort GetDisplayId();
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


Rawaho:
I might need to do some more investigation into this in the client, a few of the examples I tried didn't return the correct values.

PR #400, opened 2021-12-11
Option to generate map and table files on WorldServer build

mgasparel:
This adds msbuild parameters to the WorldServer project which allow a user to extract game table files, as well as generate base map files.

Example command which generates both the /tbl and /map folders, populated in the WorldServer's build output directory:
`dotnet build /p:"WsInstallDir=H:\Wildstar\;ExtractGameTables=true;GenerateMapFiles=true"`

I ended up adding an option to the MapGenerator to allow us to specify the output directory. I could have created then moved the files, but an output directory seemed like a reasonable addition.

I noticed that the server installation guide is on @Rawaho fork, so I didn't update the docs in this PR.

This is just a qol update to help simplify first-time setup for local development.

### DIFF
--- a/Source/NexusForever.MapGenerator/ExtractionManager.cs
+++ b/Source/NexusForever.MapGenerator/ExtractionManager.cs
@@ -18,15 +18,19 @@ namespace NexusForever.MapGenerator
             "en-GB.bin"
         };
 
+        private string outputDir;
+
         private ExtractionManager()
         {
         }
 
-        public void Initialise()
+        public void Initialise(string outputDir)
         {
             log.Info("Extracting GameTables...");
 
-            Directory.CreateDirectory("tbl");
+            this.outputDir = Path.Combine(outputDir, "tbl");
+
+            Directory.CreateDirectory(this.outputDir);
 
             ExtractGameTables();
             ExtractLanguageFiles();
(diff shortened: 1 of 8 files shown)


Source/NexusForever.MapGenerator/ExtractionManager.cs
```diff
@@ -18,15 +18,19 @@ public sealed class ExtractionManager : Singleton<ExtractionManager>
             "en-GB.bin"
         };
 
+        string outputDir;
```
Rawaho: To follow code style access modifier required.

Source/NexusForever.MapGenerator/MapGenerator.cs
```diff
@@ -41,14 +40,20 @@ private static void ParameterOk(Parameters parameters)
                 return;
             }
 
+            if ((parameters.Extract || parameters.Generate) && !string.IsNullOrEmpty(parameters.OutputDir))
```
Rawaho, line 43: Potentially default to original local directory if no explicit output directory is supplied?

Rawaho (changes requested):
Overall good PR, only a very minor change and suggestion (not required!).
I personally have never really played around with target files but I do like the idea.

Source/NexusForever.MapGenerator/MapGenerator.cs
mgasparel: It's not super intuitive, but this does actually fall back to the old behaviour if no path is specified. 

When `Path.Combine(outputDir, "tbl");` is run and `outputDir` is an empty string, it doesn't throw and you end up with a relative path like we did before this change. Tested the behavior on Windows and Linux.

Source/NexusForever.MapGenerator/ExtractionManager.cs
mgasparel: Fixed. Thoughts on generating an .editorconfig?

PR #409, opened 2022-04-26
Added customer survey

ChargeIn:
Added customer survery handling.

### DIFF
--- a/Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
+++ b/Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
@@ -98,11 +98,20 @@ namespace NexusForever.WorldServer.Network.Message.Handler
             session.EnqueueMessageEncrypted(new ServerPlayerEnteredWorld());
             session.Player.IsLoading = false;
         }
-        
+
         [MessageHandler(GameMessageOpcode.ClientCinematicState)]
         public static void HandleCinematicState(IWorldSession session, ClientCinematicState cinematicState)
         {
             session.Player.CinematicManager.HandleClientCinematicState(cinematicState.State);
+        }
+
+        /// <summary>
+        /// Client sends this when the user has filled out any customer surver. 
+        /// The response object contains the type of the survey, additional parmeters and the answers of the user.
+        /// </summary>
+        [MessageHandler(GameMessageOpcode.ClientCustomerSurveySubmit)]
+        public static void HandleClientCustomerSurvey(WorldSession session, ClientCustomerSurveySubmit surveyResponse)
+        {
         }
     }
 }
--- a/Source/NexusForever.Network.World/Message/Model/ClientCustomerSurvey.cs
+++ b/Source/NexusForever.Network.World/Message/Model/ClientCustomerSurvey.cs
@@ -8,33 +8,32 @@ namespace NexusForever.Network.World.Message.Model
     public class ClientCustomerSurveySubmit : IReadable
     {
         public SurveyType Type { get; set; }
-        // second parameter send with the survey request
         public ISurvey Survey { get; set; }
         public string Comment { get; set; }
 
         public void Read(GamePacketReader reader)
         {
-            Type = (SurveyType) reader.ReadInt(14);
+            Type = (SurveyType)reader.ReadInt(14);
 
             switch (Type)
             {
-                case SurveyType.QuestDifficulty:
+                case SurveyType.QuestGeneric:
                     Survey = new Survey.QuestDifficultySurvey();
                     break;
-                case SurveyType.QuestTSpell:
+                case SurveyType.TSpellQuest:
                     Survey = new Survey.QuestTSpellSurvey();
                     break;
-                case SurveyType.QuestHoldout:
+                case SurveyType.HoldoutQuest:
                     Survey = new Survey.QuestHoldoutSurvey();
                     break;
-                case SurveyType.Level:
+                case SurveyType.LevelUp:
                     Survey = new Survey.LevelingSurvey();
                     break;
-                case SurveyType.Challenge:
+                case SurveyType.GenericChallenge:
                     Survey = new Survey.ChallengesSurvey();
                     break;
-                // TODO figure out Bossfight/Dungeon
             }
+
             Survey.Read(reader);
 
             Comment = reader.ReadWideString();
--- a/Source/NexusForever.Network.World/Message/Model/ClientCustomerSurvey.cs
+++ b/Source/NexusForever.Network.World/Message/Model/ClientCustomerSurvey.cs
@@ -1,0 +1,43 @@
+using NexusForever.Network.Message;
+using NexusForever.Network.World.Message.Model.Shared;
+using NexusForever.Network.World.Message.Static;
+
+namespace NexusForever.Network.World.Message.Model
+{
+    [Message(GameMessageOpcode.ClientCustomerSurveySubmit)]
+    public class ClientCustomerSurveySubmit : IReadable
+    {
+        public SurveyType Type { get; set; }
+        // second parameter send with the survey request
+        public ISurvey Survey { get; set; }
+        public string Comment { get; set; }
+
+        public void Read(GamePacketReader reader)
+        {
+            Type = (SurveyType) reader.ReadInt(14);
+
+            switch (Type)
+            {
+                case SurveyType.QuestDifficulty:
+                    Survey = new Survey.QuestDifficultySurvey();
+                    break;
+                case SurveyType.QuestTSpell:
+                    Survey = new Survey.QuestTSpellSurvey();
+                    break;
+                case SurveyType.QuestHoldout:
+                    Survey = new Survey.QuestHoldoutSurvey();
+                    break;
+                case SurveyType.Level:
+                    Survey = new Survey.LevelingSurvey();
+                    break;
+                case SurveyType.Challenge:
+                    Survey = new Survey.ChallengesSurvey();
+                    break;
+                // TODO figure out Bossfight/Dungeon
+            }
+            Survey.Read(reader);
+
+            Comment = reader.ReadWideString();
+        }
+    }
+}
--- a/Source/NexusForever.Network.World/Message/Model/ServerCustomerSurvey.cs
+++ b/Source/NexusForever.Network.World/Message/Model/ServerCustomerSurvey.cs
@@ -7,12 +7,12 @@ namespace NexusForever.Network.World.Message.Model
     public class ServerNewCustomerSurveyRequest : IWritable
     {
         public SurveyType Type { get; set; }
-        public int SecParam { get; set; }
+        public uint Id { get; set; }
 
         public void Write(GamePacketWriter writer)
         {
             writer.Write(Type, 14);
-            writer.Write(SecParam, 32);
+            writer.Write(Id, 32);
         }
     }
 }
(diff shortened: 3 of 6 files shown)


MaxtorCoder:
Please follow proper C# and NF code styles, there are many examples 🙏🏼

ChargeIn:
> Please follow proper C# and NF code styles, there are many examples 🙏🏼

Sorry my bad, I adjusted the style according to the wiki.

Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
```diff
@@ -129,6 +129,8 @@ public enum GameMessageOpcode
         ClientPacked                    = 0x025C, // the same as ClientEncrypted except the contents isn't encrypted?
         ServerPlayerCreate              = 0x025E,
         ServerEntityCreate              = 0x0262,
+        ServerNewCustomerSurvey         = 0x033E,
+        ClientCustomerSurveyResponse   = 0x033F,
```
Matt-Hurd: Misaligned

Source/NexusForever.WorldServer/Network/Message/Model/ServerCustomerSurvey.cs
```diff
@@ -0,0 +1,36 @@
...
+        Bossfight       = 0x11,
+        Dungeon         = 0x12
+
+    }
+
+    [Message(GameMessageOpcode.ServerNewCustomerSurvey)]
+    public class ServerNewCustomerSurvey: IWritable
+    {
+        public SurveyType Type { get; set; }
+        public int SecParam { get; set; }
```
Matt-Hurd: This should be `Unknown1`

Source/NexusForever.WorldServer/Network/Message/Model/ServerCustomerSurvey.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public enum SurveyType
+    {
+        QuestDifficulty = 0x01,
+        QuestTSpell     = 0x02,
+        QuestHoldout    = 0x03,
+        MemError4       = 0x04,
+        MemError5       = 0x05,
+        MemError6       = 0x06,
+        Level           = 0x07, // Level as second parameter
```
Matt-Hurd: What does this comment mean?

Source/NexusForever.WorldServer/Network/Message/Model/ServerCustomerSurvey.cs
ChargeIn: The second parameter is not an unknown here, but simply the second parameter of the request. The parameter is defined by the type of the request.

E.g. if you send the client a level survey dialog, the second parameter is the level ( which will be also include in the response).

Source/NexusForever.WorldServer/Network/Message/Model/ServerCustomerSurvey.cs
ChargeIn: Depending on the type of the survey the request has a second parameter. In case of the level dialog, the second parameter is the level which will be display in the dialog.

Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
```diff
@@ -129,6 +129,8 @@ public enum GameMessageOpcode
         ClientPacked                    = 0x025C, // the same as ClientEncrypted except the contents isn't encrypted?
         ServerPlayerCreate              = 0x025E,
         ServerEntityCreate              = 0x0262,
+        ServerNewCustomerSurvey         = 0x033E,
```
kirmmin: You have Client and Server mixed up on this. This prefix is used to denote the sender of the packet. In this case, Client is sending a survey in and Server is sending a survey received response.

Source/NexusForever.WorldServer/Network/Message/Model/ClientCustomerSurvey.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class ClientCustomerSurveyResponse : IReadable
+    {
+        public SurveyType Type { get; set; }
+        // second parameter send with the survey request
+        public short SecRequestParam { get; set; } 
+
+        // Rating questions: Answers between 1 and 5
+        // Quests are dependend on the survey-type
+        // Answer can be 0 if question is not defined in the survey (mostly answer 4)
+        public short Answer1 { get; private set; }
```
kirmmin: Rather than Answer1 - Answer4, you would be better having Classes that denote the field names appropriately, then have the Reader read as a Class based on the `SurveyType`.

Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
```diff
@@ -99,5 +99,14 @@ public static void HandleClientEnteredWorld(WorldSession session, ClientEnteredW
...
             session.Player.IsLoading = false;
         }
+
+        /// <summary>
+        /// Client sends this when the user has filled out any customer surver. 
+        /// The response object contains the type of the survey, additional parmeters and the answers of the user.
+        /// </summary>
+        [MessageHandler(GameMessageOpcode.ClientCustomerSurveyResponse)]
+        public static void HandleClientCustomerSurvey(WorldSession session, ClientCustomerSurveyResponse surveyResponse)
+        {
```
kirmmin, line 114: What's the plan with this? Do you intend to store them in the database to be read by a UI?

kirmmin (changes requested):

Source/NexusForever.WorldServer/Network/Message/Model/ClientCustomerSurvey.cs
ChargeIn: While you are totally right with this, I would suggest keeping it this way for now. The original surveys will probably not been used in the way they were intendet. 

The idea was to keep the survey implementation so future content creators can use them to get feedback for new content.  Implementing all type now is probably a waste of time since most of them will never been used.

But if you still want them I can implement them. (for the sake of completeness)

Source/NexusForever.WorldServer/Network/Message/Handler/MiscHandler.cs
ChargeIn: Yes, that is the idea behind it, but implementing it now does not make sense. The idea was to implement it when people start working on new content or balancing and want feedback.

Source/NexusForever.WorldServer/Network/Message/Model/ClientCustomerSurvey.cs
kirmmin: I think we would prefer for those who wish to modify the client to also modify the server to their needs. If they're adding/removing SurveyType's, they can go so far as to add/remove a Class that supports it for the purpose of the database.

Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
ChargeIn: I guess the naming is bad here, but the prefixes are correct. The ServerNewCustomerSurvey is send by the server to the client to create the survey popup. After the user filled out the survey the client sends the ClientCustomerSurveyResponse to the server. 

Should I adjust the naming?

kirmmin: Yeah, I would reverse them:
- ServerCustomerSurveyResponse
- ClientCustomerSurveyCreate

kirmmin: `ClientCustomerSurveySubmit` is probably better than Create, so I'd go with that :P

Source/NexusForever.Shared/Network/Message/GameMessageOpcode.cs
```diff
@@ -129,6 +129,8 @@ public enum GameMessageOpcode
         ClientPacked                    = 0x025C, // the same as ClientEncrypted except the contents isn't encrypted?
         ServerPlayerCreate              = 0x025E,
         ServerEntityCreate              = 0x0262,
+        ServerCustomerSurveyResponse    = 0x033E,
```
Rawaho: When the client receives this packet it invokes a LUA function called NewCustomerSurveyRequest, could potentially name the packet around this.

Source/NexusForever.WorldServer/Network/Message/Model/ServerCustomerSurvey.cs
Rawaho: From what I can see this second parameter is the index/id, seems it is used to queue multiple surveys.
If you send this packet twice with values 0 and 1 it will show the second survey after the first is completed or ignored.
By default it seems the client supported 3? This is just what I have noticed from the quick look at the CustomerSurvey LUA code.

Source/NexusForever.WorldServer/Network/Message/Model/ServerCustomerSurvey.cs
```diff
@@ -0,0 +1,36 @@
+using NexusForever.Shared.Network;
+using NexusForever.Shared.Network.Message;
+using NexusForever.WorldServer.Network.Message.Model.Shared;
+
+namespace NexusForever.WorldServer.Network.Message.Model
+{
+    public enum SurveyType
```
Rawaho: Might want to move this to a separate file.
Usually we try to stick to a single definition (class, enum, etc...) per file unless it is nested.
In this case SurveyType is used by multiple different classes so probably should go into the static folder.

Source/NexusForever.WorldServer/Network/Message/Model/ClientCustomerSurvey.cs
```diff
@@ -0,0 +1,179 @@
...
+
+            Comment = reader.ReadWideString();
+        }
+    }
+
+    public abstract class Survey { 
+        public abstract void Read(GamePacketReader reader);
+    };
+
+    public class QuestDifficultySurvey : Survey
```
Rawaho: Same as below, might want to nest these classes.

Source/NexusForever.WorldServer/Network/Message/Model/ClientCustomerSurvey.cs
```diff
@@ -0,0 +1,179 @@
...
+                    break;
+                // TODO figure out Bossfight/Dungeon
+            }
+            Survey.Read(reader);
+
+            Comment = reader.ReadWideString();
+        }
+    }
+
+    public abstract class Survey {
```
Rawaho: Does this need to be an abstract class, would an interface do the job?
```csharp
interface ISurvey : IReadable
{
   // bla
}
```

Source/NexusForever.WorldServer/Network/Message/Model/ClientCustomerSurvey.cs
ChargeIn: Moved it to an own file and wrapped it with a Survey class. But since we cannot extract anything common among these classes, nesting does not bring any advantage.

Source/NexusForever.WorldServer/Network/Message/Model/ServerCustomerSurvey.cs
ChargeIn: As stated above the second parameter is used in some dialogs to add addition information, e.g. if you send a level survey and set the second parameter to 10 it will show ".... Level 10" in the title of the Survey. So if suggestion with the id is true than there need to be a thrid parameter.

PR #417, opened 2022-08-04
[Player/Core] Loading improvements

kirmmin:
Loading packet reorganisation based on packet logs
Player will go to loading screen immediately when transitioning maps, giving time for server to load map entities and send them
Slight fix for potential bug when switching maps


Rawaho:
This needs to be reinvestigated, going to close for the time being.

PR #448, opened 2023-05-07
[Map/Core] Fix grid count variables overflowing

kirmmin:
This leads to map lockups where the grids freeze in unloading state. While this doesn't solve the problem in entirety, the grid should at least unload when it reaches 0 entities now, instead of freezing.

### DIFF
--- a/Source/NexusForever.Game/Map/MapGrid.cs
+++ b/Source/NexusForever.Game/Map/MapGrid.cs
@@ -128,7 +128,10 @@ namespace NexusForever.Game.Map
             if (UnloadStatus.HasValue)
                 return;
 
-            visibilityCount++;
+            checked
+            {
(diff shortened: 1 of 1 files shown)


Fabi:
This will trigger exceptions to control the flow after overflows occur. That feels very wrong to do...

kirmmin:
This is not a solution to the initial problem, but a way of capturing the problem happening. Right now, the grids just freeze up. It's also our practice to use `checked` across unsigned values being modified in this way across the codebase, so if nothing else it brings it in line with everything else.

Rawaho:
I do agree this isn't a "fix" but it is probably good to have regardless.
As mentioned it brings code inline with what is done elsewhere and also causes a fail fast scenario which I would prefer :D

PR #449, opened 2023-05-09
[Entity/Vehicle] Fix vehicle attach for players mounted already when they enter range

kirmmin:
The following issue occurs occasionally when entering world near someone who is already mounted. The packet adjustments made in this PR resolve the problem and also updates our packet flow to match live packet logs.
This is currently more prevalent on my test server, but I believe that is because I match live by sending `0x083B` after every non-Player entity's `ServerEntityCreate` packet, like live packets did.

image

### DIFF
--- a/Source/NexusForever.Game/Entity/Player.cs
+++ b/Source/NexusForever.Game/Entity/Player.cs
@@ -663,11 +663,33 @@ namespace NexusForever.Game.Entity
                 Session.EnqueueMessageEncrypted(worldEntity.BuildCreatePacket());
 
             if (entity is IPlayer playerEntity)
+            {
+                // Inform this player of visible player's path. This appears to be just used by the tooltip system.
                 Session.EnqueueMessageEncrypted(new ServerSetUnitPathType
                 {
                     Guid = playerEntity.Guid,
                     Path = playerEntity.Path
                 });
+
+                // Send a packet to "connect" the player to their vehicle.
+                // This is more prevalent of a problem when we use 0x08B3 after every non-player entity's ServerEntityCreate packet like live packets did.
+                if (playerEntity.VehicleGuid > 0)
+                {
+                    IVehicle vehicle = Map.GetEntity<IVehicle>(playerEntity.VehicleGuid);
+                    if (vehicle != null)
+                    {
+                        IVehiclePassenger passenger = vehicle.ToList().SingleOrDefault(p => p.Guid == playerEntity.Guid);
+                        if (passenger != null)
+                            Session.EnqueueMessageEncrypted(new ServerVehiclePassengerSet
+                            {
+                                Self = passenger.Guid,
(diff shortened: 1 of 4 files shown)


Rawaho:
This probably needs to be handled differently.
Vehicle attachment code really should sit with the vehicle implementation and not elsewhere.

kirmmin:
This packet is only sent when a player entity becomes visible to other players after the vehicle has been created and already attached. This is more like a "Hey, Player B. Player A is already attached to Vehicle A, just fyi. See ya 👊", lol

Our practice has been to place this in AddVisible so far. Do you have a suggestion on where this should go?

Rawaho:
Would it not make more sense for it to be included when we send the vehicle itself to the other clients?
Like, create vehicle 123, attach player to vehicle 123, one after the other?
We might need to consider order in such case to insure the player is created first, followed by the vehicle and then the attach?

kirmmin:
We already do that. Right now, we're sending 2 packets (`ServerVehiclePassengerSet` and `ServerVehiclePassengerAdd`). They do the same thing, but have 2 specific usecases based on packets. 
The former is used ONLY when a Player is already mounted and another Player has them come into range. The second player requires this packet to do the same `ServerVehiclePassengerAdd` but not at "vehicle spawn" time.

kirmmin:
If I misunderstood and you were suggesting when we AddVisible for the vehicle, the packet logs have `ServerVehiclePassengerSet` coming after the Player every time - they also sent the mount for the Player before the Player packet, but we don't currently have ordering.

Dekker3D:
Just a thought: maybe have a second AddEntity-style call to run after all newly-visible entities have actually been added, so you can be quite certain that both the player and mount have been described to the client already?

Or, after a brief skim through this commit: it's in AddVisible. Maybe only call the AddVisible function after all newly-visible entities have been added, rather than calling it after adding each individual entity?

Source/NexusForever.Game/Entity/Player.cs
```diff
@@ -663,12 +663,34 @@ public override void AddVisible(IGridEntity entity)
...
                 });
 
+                // Send a packet to "connect" the player to their vehicle.
+                // This is more prevalent of a problem when we use 0x08B3 after every non-player entity's ServerEntityCreate packet like live packets did.
+                if (playerEntity.VehicleGuid > 0)
+                {
+                    IVehicle vehicle = Map.GetEntity<IVehicle>(playerEntity.VehicleGuid);
+                    if (vehicle != null)
+                    {
+                        IVehiclePassenger passenger = vehicle.ToList().First(p => p.Guid == playerEntity.Guid);
```
MaxtorCoder:
```suggestion
                        IVehiclePassenger passenger = vehicle.ToList().FirstOrDefault(p => p.Guid == playerEntity.Guid);
```

`First` will throw an exception when the `playerEntity.Guid` is not found in the container.

PR #450, opened 2023-05-25
[Entity/Properties]: Property Updates

kirmmin:
- Player and Class Base Properties
    - These are stored in a DB table to capture data that weas on live, but also allow for customisation.
- Item Properties
    - Supports Innate and Budgeted Properties.
    - @Rawaho went to great lengths to get this reversed. (Excellent job!)
- Spell UnitPropertyModifier Effects
    - Rewrote this in entirety. We are now calculating data much closer to live. 
    - There is still some unknowns, but this implementation is much easier to understand and extend.

Rebased for `game_rework` branch.

Replaces #179


Source/NexusForever.Game/Entity/WorldEntity.cs
```diff
@@ -239,20 +304,251 @@ public virtual void OnActivateCast(IPlayer activator)
...
+            
+            ServerEntityPropertiesUpdate propertyUpdatePacket = new ServerEntityPropertiesUpdate()
+            {
+                UnitId = Guid
+            };
+            
+            foreach (Property propertyUpdate in DirtyProperties)
+            {
+                IPropertyValue propertyValue = CalculateProperty(propertyUpdate);
+                if (Properties.ContainsKey(propertyUpdate))
```
Fabi: You can skip this check and just do a `Properties[propertyUpdate] = propertyValue;`

Source/NexusForever.Game/Entity/WorldEntity.cs
```diff
@@ -239,20 +304,251 @@ public virtual void OnActivateCast(IPlayer activator)
...
+        }
+
+        /// <summary>
+        /// Add a <see cref="Property"/> modifier given a Spell4Id and <see cref="PropertyModifier"/> instance
+        /// </summary>
+        public void AddItemProperty(Property property, ItemSlot itemSlot, float value)
+        {
+            if (ItemProperties.ContainsKey(property))
+            {
+                var itemDict = ItemProperties[property];
```
Fabi: TryGetValue would save one lookup

PR #451, opened 2023-05-25
[Script/Core] Fix crashes due to null scriptCollection

kirmmin:


Fabi:
Why are these null anyways? They should never be unless disposed (on server shutdown). Do they not default to empty collections if no scripts are available?

Rawaho:
I'm going to merge this to fix the immediate issue and align it with other script owners.
I do want to investigate though whether or not an empty script collection should be provided by default.

PR #453, opened 2023-07-30
Fixed pet summoning

ChargeIn:
Fixed a null pointer on pet summon due to the id not being set.


Rawaho:
You don't need to specify `SetVisualEmit(false);` anymore, was updated in a recent commit, safe to remove.
You'll also need to add the corresponding methods to the interface IWorldEntity.

ChargeIn:
Since the original initialize is also not in the interface, should I add it as well?

Rawaho:
Honestly the old method can be removed and be replaced with Initialise and SetVisualInfo.
Eventually I expect there to be a `Initialise(ICreatureInfo)` or something to replace.

Rawaho:
Looks good, thank you for the PR and contribution 👍

PR #454, opened 2023-08-05
Added pet renaming

ChargeIn:
Added support for renaming the pets.

Since testing might be a bit tricky since the scanbot logic is still missing you can use the attached patch file to allow deploying the scan bot of a scientist and rename it.

[scan_bot_deploy.patch](https://github.com/NexusForever/NexusForever/files/12346582/scan_bot_deploy.patch)

### DIFF
--- a/Source/NexusForever.Game.Abstract/Entity/IPetCustomisationManager.cs
+++ b/Source/NexusForever.Game.Abstract/Entity/IPetCustomisationManager.cs
@@ -11,6 +11,11 @@ namespace NexusForever.Game.Abstract.Entity
         void UnlockFlair(ushort id);
 
         /// <summary>
+        /// Renames the pet name for <see cref="PetType"/> and object id.
+        /// </summary>
+        void RenamePet(PetType type, uint objectId, String name);
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


Rawaho:
It would be good if the rename validation could also be checked as part of this PR.
I assume this system already partially exists via UserText but would be good to check.

ChargeIn:
I added the same validation as used by the client. The only thing missing is the profanity filter. But since I'm not really familiar with the policy regarding external libs, I left a todo for now.

Rawaho:
This is actually already built into the core, see TextFilterManager and UserText.
It handles both text profanity and validity, it is based on how the client does it.

It might need to be expanded to handle new rules though as not all of them are handled.

ChargeIn:
Ah nice, 👍.  I will be away for a week afterwards i will look into adding the missing logic.

Rawaho:
image
When you get to it, these seem to be the values in the client.

ChargeIn:
Thx for the values. I tried to look into the client logic validation, but it seems a bit to much for me for now 😅.

I matched the values you provided, squashed my commits and cleaned up the branch.

PR #461, opened 2023-10-15
Place-from-vendor plotindex fix

### LINKED ISSUE #218: [Bug] Housing - Decor - Placing decor not already in crate causes large offset

Dekker3D:
Should fix https://github.com/NexusForever/NexusForever/issues/218; the PlotIndex wasn't assigned to the decor when the client placed decor directly from vendor.

### DIFF
--- a/Source/NexusForever.Game/Map/Instance/ResidenceMapInstance.cs
+++ b/Source/NexusForever.Game/Map/Instance/ResidenceMapInstance.cs
@@ -355,6 +355,7 @@ namespace NexusForever.Game.Map.Instance
 
             IDecor decor = residence.DecorCreate(entry);
             decor.Type = update.DecorType;
+            decor.PlotIndex = update.PlotIndex;
 
             if (update.ColourShiftId != decor.ColourShiftId)
             {


Dekker3D:
Updated for current rework branch.

PR #468, opened 2024-01-31
Fix linux build

BenjaminCDP:
Summary :
* Fix all Nexus.Forever.Script.* .csprojs files to allow building on linux/OSX.
* Fix SourceFinder.cs to be OS-agnostic.

I'd like to write the Server installation guide for Linux if possible.


Fabi:
I think a better general solution would be to simply set the output folder on the projects. Avoids any manual copying for those scripts

Rawaho:
@BenjaminCDP thank you for contribution, this has been merged :)

PR #472, opened 2024-05-29
Corrected TargetGroupId passed to Quest.ObjectiveUpdate in RewardKill

Infinite-Pathways:
* RewardKiller was passing CreatureId instead of TargetGroupId in UnitEntity.RewardKiller.
* Correction applied using GameTable data.


Rawaho:
Good catch! Thank you for the pull request :D

PR #473, opened 2024-05-30
Message handler for client packet 0x23C

krakal:
This packet a couple unrelated sets of data.
First is a random number the client generates and its seed value, second is a set of statistics on the watchdog task which gives an indication of how much load the client's computer is under. There is currently nothing done with the data when received but the handler is there to stop the unknown packet log messages.


Source/NexusForever.Network.World/Message/Model/ClientWatchdogStatistics.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Net.Sockets;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NexusForever.Network.World.Message.Model
+{
+    [Message(GameMessageOpcode.ClientWatchdogStatistics)]
+    public class ClientWatchdogStatistics : IReadable
+    {
+        public ulong randomValueSeed { get; private set; }
```
Rawaho: Public properties should be Pascal case

Source/NexusForever.Network.World/Message/Model/ClientWatchdogStatistics.cs
```diff
@@ -0,0 +1,34 @@
+﻿using NexusForever.Network.Message;
+using NLog.Fluent;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading.Tasks;
```
Rawaho: Please remove unnecessary includes.

Rawaho (changes requested):
Thank you for the pull request.
Looks good, just a few style issues before I can merge.

Rawaho (approved):

PR #474, opened 2024-05-30
Addition of server command to allow changing client's entity send update interval

krakal:
Figured out what the unknown parameter in server message 0xF1 (ServerInstanceSettings) does. It sets the client's maximum send update interval for the player's unit. This same parameter can be changed at any time by sending server message 0x70 (ServerClientEntityUpdateInterval).

Based on captures, the server would always set this parameter as 125 or 126ms and would periodically change it but it is not clear why.


Rawaho:
Interesting find, thank you for the pull request :)

PR #480, opened 2024-08-21
Fix: Appearance changes don't stick sometimes

Dekker3D:
Fixes a bug where appearance changes don't stick sometimes, because of they're treated as "default" and not committed to the database. This PR does not include a database migration.

### DIFF
--- a/Source/NexusForever.Database.Character/CharacterContext.cs
+++ b/Source/NexusForever.Database.Character/CharacterContext.cs
@@ -379,13 +379,11 @@ namespace NexusForever.Database.Character
 
                 entity.Property(e => e.Id)
                     .HasColumnName("id")
-                    .HasColumnType("bigint(20) unsigned")
-                    .HasDefaultValue(0);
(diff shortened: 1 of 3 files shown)


Rawaho:
Is the default change required for the id too?
I would assume it is unlikely for a character to have an id of 0, although for the bone index can be 0 and not commited to the database.

Dekker3D:
> Is the default change required for the id too? I would assume it is unlikely for a character to have an id of 0, although for the bone index can be 0 and not commited to the database.

I'll be honest, I didn't check. I've run into so many issues with HasDefaultValue, on things where it makes no sense to leave that field blank, that I kinda just assume it's better not to have it in those places (so, specifically in any place where the field should never be left blank anyway). I'd recommend doing the same all over the place, tbh, unless there's a good reason to have them?

Rawaho:
I think you are probably correct, a larger look at some of the database schema is probably required.
Happy to just merge the changes to these two columns and I'll take a look if this can be done across the board.

PR #481, opened 2024-08-28
Fix incorrect filename generation on Linux for map files

Kinsmir:
### Problem:

When generating map files on Linux, the filenames are incorrectly generated as `Map\<Filename>`. This is due to platform-specific path delimiters being used in the filename retrieval.

Screenshot_2024-08-28_09-49-35

### Workaround:

This pull request introduces a fix for getting the filename on Linux.

### Testing:

While the changes have been thoroughly tested on Linux, additional testing on Windows is recommended to verify that the fix does not introduce unintended side effects on that platform.

### DIFF
--- a/Source/NexusForever.MapGenerator/GenerationManager.cs
+++ b/Source/NexusForever.MapGenerator/GenerationManager.cs
@@ -69,7 +69,7 @@ namespace NexusForever.MapGenerator
         /// </summary>
         private void ProcessWorld(WorldEntry entry, byte? gridX = null, byte? gridY = null)
         {
-            var mapFile = new WritableMapFile(Path.GetFileName(entry.AssetPath.Replace('\\', '/')));
+            var mapFile = new WritableMapFile(Path.GetFileName(entry.AssetPath.Replace('\\', Path.DirectorySeparatorChar)));
 
             log.Info($"Processing {mapFile.Asset}...");
 
--- a/Source/NexusForever.MapGenerator/GenerationManager.cs
+++ b/Source/NexusForever.MapGenerator/GenerationManager.cs
@@ -69,7 +69,7 @@ namespace NexusForever.MapGenerator
         /// </summary>
         private void ProcessWorld(WorldEntry entry, byte? gridX = null, byte? gridY = null)
         {
-            var mapFile = new WritableMapFile(Path.GetFileName(entry.AssetPath));
+            var mapFile = new WritableMapFile(Path.GetFileName(entry.AssetPath.Replace('\\', '/')));
 
             log.Info($"Processing {mapFile.Asset}...");
 


Fabi:
That moment when someone just posts a ChatGPT output...Don't do that

That being said .Replace('\\', '/'); should just be appended to the filepath and that's it.

Dekker3D:
Omechron does something similar. Just replacing the \ with / is enough.

Kinsmir:
> That moment when someone just posts a ChatGPT output...Don't do that

I only asked ChatGPT to rephrase partial my pull request message, this is my normal style for pull requests.

> That being said .Replace('\', '/'); should just be appended to the filepath and that's it.

I didn't want to mess with it too much, as the files generated were correct. But yes, you can append the replace directly too.

Fabi:
No. not there. u gotta do that on the filepath that gets saved

Kinsmir:
> No. not there. u gotta do that on the filepath that gets saved

Changing the line here won't give the correct result. See the accompanying image.

image

I also found out that it went wrong much earlier, which is why I changed the line in the commit. Because we are trying to get the filename of the map as the very first step of this function. But because of the wrong path, it doesn't get the correct filename on Linux, and instead gets the whole path as filename, so it ends up with the wrong write path.

This is not a problem on Windows, where the path issue does not apply. So by making sure we get the correct filename first. We don't need to change anything in the write section later.

So my first attempt and assumption based on behaviour was wrong. But after investigating the cause shown above, I found the core problem.

MaxtorCoder:
Not there either… Literally after `Path.Combine` you add the `Replace`… Not on the `mapFile.Asset` property

Fabi:
If the issue already exists for reading the file then this change is fine.

I wonder if there are more places in the source where path separators are wrong. Can you look for those and see if you can add them?

Dekker3D:
Ah, it seems I misread. This change is in the mapgenerator, while the change I made in Omechron was in BaseMapManager. Mine is an 8 line function that's equivalent to Path.Combine(assetPath.Split('\\')), where assetPath is the asset path in LoadBaseMap(string assetPath). I'll submit a PR for it or something. I don't even bother running the mapgenerator on Linux so I hadn't noticed that the mapgenerator has the same issue.

Kinsmir:
> If the issue already exists for reading the file then this change is fine.
> 
> I wonder if there are more places in the source where path separators are wrong. Can you look for those and see if you can add them?

I did not notice anything immediately when testing and/or running NexusForever after generating the map files. At first glance all files seem to load and write correctly. But I will try to see if there are other instances in the code where this remains a problem.

Rawaho:
Reading was fixed awhile ago in commit 18dfe79cdecf0f6e8141d03aad83b5d56fcf302c

I guess this slipped through the cracks since most generate the map on Windows still 😄

Kinsmir:
> Reading was fixed awhile ago in commit [18dfe79](https://github.com/NexusForever/NexusForever/commit/18dfe79cdecf0f6e8141d03aad83b5d56fcf302c)
> 
> I guess this slipped through the cracks since most generate the map on Windows still 😄

I haven't found any other cases in the code. But I did find, as your comment just shows, that there is also a proper const for the directory separator. So I updated the PR accordingly.

PR #483, opened 2024-08-30
Update PrerequisiteTypes

Kinsmir:
After going through the long list of enums for `PrerequisiteTypes` and game translations from English, German and French. Numbers 1 to 44 are based on the enum value instead of the error message.

This is the list of game messages and types I have managed to map so far.

There are a few names here that I'm not entirely happy with yet, but we need a bit more context for what they do exactly.


Rawaho:
Thank you for the effort in collating these types.
While some of the names might not be ideal at the moment, it is a great starting point.

I have rebased the PR to the latest game_rework and merged.

PR #487, opened 2025-03-11
fix: Move the character packet generation to a manager

jasoncouture:
This was noted in a TODO comment, and in another branch of mine was getting in the way. This may not be the perfect cleanup, but it should be a good enough start.


Rawaho:
Thank you for the contribution, much better than what was there previously.

PR #492, opened 2025-04-19
P2P Trading

krakal:
*Added all messages involved with peer-to-peer trading.

Still some unknown fields in the transmitted trade item data.


Source/NexusForever.Network.World/Message/Model/ServerP2PTradeResult.cs
```diff
@@ -0,0 +1,40 @@
...
+            TargetNotAllowedToTrade = 17,
+            InviteFailedNoTrading = 18
+        }
+
+        public P2PTradeResult Result { get; set; }
+        public bool Cancelled { get; set; }
+
+        public void Write(GamePacketWriter writer)
+        {
+            writer.Write(Result);
```
Rawaho: Assuming this is a typo and should be ``writer.Write(Result, 5)``?
By default an enum writes as 64 bits, quick check in the client seems to show 5 bits.

Rawaho (changes requested):
Great contribution, only what I believe is a small typo needs to be resolved.

Source/NexusForever.Network.World/Message/Model/ServerP2PTradeResult.cs
krakal: Good catch.

PR #493, opened 2025-05-01
Crafting/tradeskill messages

krakal:
*Adds all crafting messages but no handlers.


Source/NexusForever.Network.World/Message/Model/ClientCraftingRuneSlotAdd.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public ulong ItemGuid { get; private set; }
+        public bool IsNotFusion { get; private set; }
+        public RuneType Type { get; private set; }
+
+        public void Read(GamePacketReader reader)
+        {
+            ItemGuid = reader.ReadULong();
+            IsNotFusion = reader.ReadBit();
+            Type = (RuneType)reader.ReadByte(5);
```
Rawaho: Just to keep things consistent, should use ReadEnum here.
This should be done for all examples in the PR.

Source/NexusForever.Network.World/Message/Model/ServerCraftingDiscoveryHotCold.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class ServerCraftingDiscoveryHotCold : IWritable
+    {
+        public bool Pass { get; set; }
+        public uint TradeskillSchematic2IdCrafted { get; set; }
+        public uint Item2IdCrafted { get; set; }
+        public uint Unused { get; set; }
+        public CraftingDiscovery HotOrCold { get; set; }
+        public CraftingDirection Direction { get; set; }
+        public uint EarnedXp { get; set; }
+        public List<uint> MaterialReturnedIds { get; set; } = new List<uint>();
```
Rawaho: Could use new initialiser syntax here `List<uint> MaterialReturnedIds { get; set; } = []`
This should be updated for examples in PR.

Source/NexusForever.Network.World/Message/Model/ServerProfessionModifiers.cs
```diff
@@ -0,0 +1,40 @@
...
+    public class ServerProfessionModifiers : IWritable
+    {
+        public class CraftingModifier : IWritable
+        {
+            CraftingModifierType Type { get; set; }
+            TradeskillType TradeskillId { get; set; }
+            uint Item2TypeId { get; set; }
+            uint Item2MaterialId { get; set; }
+            float Coefficient { get; set; } // Might be used additive or multiplier, depends on Type
+            uint FixedValue { get; set; }
```
Rawaho: Missing public keyword.

Source/NexusForever.Network.World/Message/Model/ServerProfessionsLoad.cs
```diff
@@ -0,0 +1,58 @@
...
+                writer.Write(TradeskillSchematic2Id);
+                writer.Write(Coordinates.X);
+                writer.Write(Coordinates.Y);
+            }
+        }
+
+        List<TradeskillInfo> Tradeskills { get; set; } = new List<TradeskillInfo>();
+        List<uint> LearnedSchematics { get; set; } = new List<uint>(); // TradeskillSchematic2Id
+        List<DiscoveredSchematic> DiscoveredSchematics { get; set; } = new List<DiscoveredSchematic>();
+        List<uint> UnknownArray { get; set; } = new List<uint>();
```
Rawaho: Missing public keyword.

Source/NexusForever.Network.World/Message/Model/Shared/CraftStats.cs
```diff
@@ -0,0 +1,55 @@
...
+            temp |= (ulong)ApSpSplit;
+            temp <<= 8;
+
+            temp |= (ulong)Unknown2;
+            temp <<= 8;
+
+            writer.Write(temp);
+        }
+    }
+}
```
Rawaho, line 55: What is the reason for reading and writing the packet this way?
The function used in the client just reads and writes a list of bytes, this usually is done when a union or array is used at the client.
Was there anything like this?

Source/NexusForever.Network.World/Message/Model/Shared/Crafting.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        Air     = 0x7,
+        Fire    = 0xA,
+        Water   = 0x8,
+        Earth   = 0x9,
+        Logic   = 0xB,
+        Life    = 0xC,
+        Fusion  = 0xD,
+    };
+}
```
Rawaho: Some of these should probably go into Game.Static eventually but fine having it in the network library for the time being.
Would split into seperate file though.

Rawaho (changes requested):
Thank you for the contribution, just a few nitpicks and a question around the reading and writing of the CraftStats.

Source/NexusForever.Network.World/Message/Model/Shared/CraftStats.cs
krakal: This is a strange one. The packet read is straightforward but the handlers do this extra byte packing/unpacking step. I opted to do the byte packing/unpacking at the message rather than in the handler. As you can see by the unknowns, I still don't really understand all the data, but the data looks sensible when I parse sniffs with it.

Source/NexusForever.Network.World/Message/Model/ClientTradeskillLearn.cs
```diff
@@ -0,0 +1,18 @@
...
+    [Message(GameMessageOpcode.ClientTradeskillLearn)]
+    public class ClientTradeskillLearn : IReadable
+    {
+        public TradeskillType ToLearnTradeskillId { get; private set; }
+        public TradeskillType ToDropTradeskillId { get; private set; }
+
+        public void Read(GamePacketReader reader)
+        {
+            ToLearnTradeskillId = reader.ReadEnum<TradeskillType>();
+            ToDropTradeskillId = reader.ReadEnum<TradeskillType>();
```
Rawaho: Assuming both of these should be `reader.ReadEnum<TradeskillType>(32)`?
Default is 64 bits on ReadEnum.

Source/NexusForever.Network.World/Message/Model/ClientTradeskillPickTalent.cs
```diff
@@ -0,0 +1,20 @@
...
+    [Message(GameMessageOpcode.ClientTradeskillPickTalent)]
+    public class ClientTradeskillPickTalent : IReadable
+    {
+        public TradeskillType TradeskillId { get; private set; }
+        public uint Tier { get; private set; }
+        public uint TradeskillBonusId { get; private set; }
+
+        public void Read(GamePacketReader reader)
+        {
+            TradeskillId = reader.ReadEnum<TradeskillType>();
```
Rawaho: Same here, 32 bits?

Source/NexusForever.Network.World/Message/Model/ClientTradeskillResetTalents.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace NexusForever.Network.World.Message.Model
+{
+    [Message(GameMessageOpcode.ClientTradeskillResetTalents)]
+    public class ClientTradeskillResetTalents : IReadable
+    {
+        public TradeskillType TradeskillId { get; private set; }
+
+        public void Read(GamePacketReader reader)
+        {
+            TradeskillId = reader.ReadEnum<TradeskillType>();
```
Rawaho: Same here, 32 bits?

Source/NexusForever.Network.World/Message/Model/ServerAddLearnedSchematic.cs
```diff
@@ -0,0 +1,22 @@
...
+    [Message(GameMessageOpcode.ServerAddLearnedSchematic)]
+    public class ServerAddLearnedSchematic : IWritable
+    {
+        public TradeskillType TradeskillId { get; set; }
+        public uint TradeskillSchematic2Id { get; set; }
+        public Vector2 DiscoveryCoordinates { get; set; } // Coordinates on the discovery panel
+
+        public void Write(GamePacketWriter writer)
+        {
+            writer.Write(TradeskillId);
```
Rawaho: Same here, 32 bits?

Source/NexusForever.Network.World/Message/Model/ServerRemoveLearnedSchematic.cs
```diff
@@ -0,0 +1,17 @@
+using NexusForever.Network.Message;
+
+namespace NexusForever.Network.World.Message.Model
+{
+    [Message(GameMessageOpcode.ServerRemoveLearnedSchematic)]
+    public class ServerRemoveLearnedSchematic : IWritable
+    {
+        public uint TradeskillId { get; set; }
```
Rawaho: Is this value the same as the enum TradeskillType?

Rawaho (changes requested):
Looks good, just a few enums I assume need to be 32 bits.

Rawaho:
Thank you for the contribution :D

PR #494, opened 2025-05-11
Matchmaking additions and corrections

krakal:
* Adds all missing matchmaking messages.
* Adds or corrects information in some existing matchmaking messages.


Source/NexusForever.Game/Matching/Match/MatchProposal.cs
```diff
@@ -137,17 +137,17 @@ private void SendMatchReady(bool inProgress, IMatchProposalTeam allyTeam)
...
             }
             else
             {
                 message = new ServerMatchingMatchReady()
                 {
                     MatchType    = MatchingQueueGroup.MatchType,
-                    TotalAllies  = allyTeam.MemberCount,
-                    TotalEnemies = enemyTeam?.MemberCount ?? 0u,
+                    UnacceptedAllies  = allyTeam.MemberCount, // Needs works, as above
+                    UnacceptedEnemies = enemyTeam?.MemberCount ?? 0u, // Needs works, as above
```
Rawaho: I'm not sure if this really needs any change.
Currently there is no scenario where a player will get the ServerMatchingMatchReady packet when another player has accepted.
The pending allies and enemies will always be the entire team.
You are correct though, the client calls this PendingAllies/PendingEnemies.

Source/NexusForever.Game/Matching/Match/MatchProposal.cs
```diff
@@ -137,17 +137,17 @@ private void SendMatchReady(bool inProgress, IMatchProposalTeam allyTeam)
                 message = new ServerMatchingMatchInProgressReady()
                 {
                     MatchType    = MatchingQueueGroup.MatchType,
-                    TotalAllies  = allyTeam.MemberCount,
-                    TotalEnemies = enemyTeam?.MemberCount ?? 0u,
+                    AcceptedAllies  = allyTeam.MemberCount, // Needs concept of total invited allies, accepted allies, and unaccepted allies
+                    UnacceptedAllies = 0, // Related to the above
```
Rawaho: Client calls this PendingAllies and CurrentAllies.

Source/NexusForever.Network.World/Message/Model/ServerMatchingListPlayersQueuedForMap.cs
```diff
@@ -0,0 +1,88 @@
+﻿using NexusForever.Game.Static.Reputation;
+using NexusForever.Network.Message;
+using NexusForever.Network.World.Message.Model.Shared;
+
+namespace NexusForever.Network.World.Message.Model
+{
+    [Message(GameMessageOpcode.ServerMatchingListPlayersQueuedForMap)]
```
Rawaho, line 7: This looks like a debug/admin related packet?

Source/NexusForever.Network.World/Message/Model/ServerMatchingPvpRatingUpdated.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace NexusForever.Network.World.Message.Model
+{
+    [Message(GameMessageOpcode.ServerMatchingPvpRatingUpdated)]
+    public class ServerMatchingPvpRatingUpdated : IWritable
+    {
+        public uint Rating { get; set; }
+        public uint Wins { get; set; }
+        public uint Losses { get; set; }
+        public uint Draws { get; set; }
+        MatchingGameRatingType Type { get; set; }
```
Rawaho: Should be public.

Source/NexusForever.Network.World/Message/Model/ServerMatchingRoleCheckStarted.cs
```diff
@@ -5,7 +5,7 @@ namespace NexusForever.Network.World.Message.Model
     [Message(GameMessageOpcode.ServerMatchingRoleCheckStarted)]
     public class ServerMatchingRoleCheckStarted : IWritable
     {
-        public bool RolesRequired { get; set; }
+        public bool RolesRequired { get; set; } // Passed to Lua event MatchingRoleCheckStarted but not used in Carbine's Lua code
```
Rawaho: This determines whether or not to show the option to choose a role, see MatchMaker:OnRoleCheck(bRolesRequired).
I believe this is used for the party leader since they choose roles when queuing but the others in the party haven't.

Source/NexusForever.Network.World/Message/Model/ServerPvpRatingUpdate.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace NexusForever.Network.World.Message.Model
+{
+    [Message(GameMessageOpcode.ServerPvpRatingUpdate)]
+    public class ServerPvpRatingUpdate : IWritable
+    {
+        public class PvpRating : IWritable
+        {
+            public uint Category { get; set; } // Arena2v2, Arena3v3, Arena5v5, RatedBattleground = 5, Warplot = 1
+            MatchingGameRatingType Type { get; set; }
```
Rawaho: Public.

Source/NexusForever.Network.World/Message/Model/ServerPveRatingUpdate.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    // Very similar to ServerPvpRatingUpdate, but not sure what goes in it
+    // This triggers the lua event PveRatingUpdated but there are no uses of it in Carbine's lua code
+    [Message(GameMessageOpcode.ServerPveRatingUpdate)]
+    public class ServerPveRatingUpdate : IWritable
+    {
+        public class PveRating : IWritable
+        {
+            public uint Category { get; set; } // similar to ServerPvpRatingUpdate, not sure what the categories are
+            MatchingGameRatingType Type { get; set; }
```
Rawaho: Public.

Rawaho (commented):
Looking good, just a couple of things.

Source/NexusForever.Network.World/Message/Model/ServerMatchingListPlayersQueuedForMap.cs
krakal: Correct. It was there amongst the other matchmaking messages so I thought why not add it.

Source/NexusForever.Network.World/Message/Model/ServerMatchingRoleCheckStarted.cs
krakal: Yes, the comment is just pointing out that the OnRoleCheck function doesn't actually do anything with the bRolesRequired variable passed to it.

The packet does set the roleCheckActive flag for the player's client which then allows them to send role check messages, but the C++ code doesn't use the value of RolesRequired for this.

Source/NexusForever.Game/Matching/Match/MatchProposal.cs
krakal: I think this depends on if you allow for matchmaking to deal with not having 100% queue for both teams.
For example, if a match of 10 versus 10 gets made, if one person drops immediately and there is an available person in the queue while some of the other players have accepted, the server might then bring in the available person from the queue and they would need to be sent an up-to-date number of pending players before getting the incremental updates.

Source/NexusForever.Game/Matching/Match/MatchProposal.cs
krakal: Worth mentioning that the lua code calls the value AcceptedAllies. TotalAllies is used in the lua code, but this is derived from CurrentAllies + PendingAllies in the C++.

Source/NexusForever.Network.World/Message/Model/ServerMatchingRoleCheckStarted.cs
Rawaho: The LUA function does use the variable on the last line? To show or hide the UI element?

krakal: image
The LuaEvent call just passes the variable value to the lua event.

krakal: In the lua side, the bRoleRequired argument is not used.
image

krakal: Or unless the Show implicitly uses the function argument?

krakal: Ah, more than one event handler for this event. I was looking at the MatchMaker.lua handler for that event.

This is for the ConfirmRole.lua:

image

PR #496, opened 2025-05-20
Quests - tweaks and additions

krakal:
* Corrections to some names in a few existing quest messages
* Added a number of quest and quest related messages (dialog, phase visibility)


Source/NexusForever.Network.World/Message/Model/ClientQuestComplete.cs
```diff
@@ -7,13 +7,13 @@ public class ClientQuestComplete : IReadable
     {
         public ushort QuestId { get; private set; }
         public ushort RewardSelection { get; private set; }
-        public bool IsCommunique { get; private set; }
+        public bool IsCommunicatorMsg { get; private set; }
```
Rawaho, line 10: Client calls this IsCommunique, if we think the original name is bad happy for it to change, although where possible it should match the client if we have it.

Source/NexusForever.Network.World/Message/Model/ServerQuestLuaEvent.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            writer.Write(LuaEventId);
+            writer.Write(LuaEvents.Count);
+            foreach (var luaEvent in LuaEvents)
+            {
+                luaEvent.Write(writer);
+            }
+        }
+    }
+}
```
Rawaho, line 92: I would probably update this to use an interface, similar to how ServerCombatLog is handled.
In general though, this LuaEvent is interesting, not something I had noticed before in the client, good find :)

Rawaho (changes requested):
A good PR, just two comments.

Source/NexusForever.Network.World/Message/Model/ClientQuestComplete.cs
krakal: The client doesn't have a name for this. This is the name I came up with while reverse engineering as it made sense. IsCommunique is ok but my brain likes to read it as IsComm-unique as there is no accent mark on it. Also, all the tbl data is called communicatorMessages.

Source/NexusForever.Network.World/Message/Model/ServerQuestLuaEvent.cs
krakal: It started as an interface and changed to a class so the derived classes didn't have to implement the Type property as it could just be inherited.

krakal: @Rawaho Any further thoughts on this?  I don't expect we will ever want other semi/un-related classes to implement these properties so it does not feel like an interface is appropriate.

PR #501, opened 2025-06-07
PublicEvent message - additions and corrections

krakal:
* Adds remaining messages for PublicEvents and fixes a few existing messages
* Moves PublicEvent messages to their own folder as part of ongoing tidying of Message\Model directory
* Adjustment of existing Event namespaces to PublicEvent to make things line up across the projects


Rawaho:
Not sure if it makes sense to standardise the namespaces? Network and static is now PublicEvent but everything in Game is still Event.

krakal:
I only touched the namespace in areas I was in at the time, not across every project, as I was trying to keep the number of files touche by the PR to a reasonable number. My intention was to update the namespace in other the other projects if this gets merged.

krakal:
If you would like, I can update this PR so the namespaces in all the projects are consistent.

Rawaho:
I do agree keeping it small, although in that scenario I wouldn't have made any namespace changes and just done them all as a seperate PR.
Happy to change it across the board, better to have it consistent imo

Source/NexusForever.Game.Static/PublicEvent/PublicEventStat.cs
```diff
@@ -16,10 +16,15 @@ public enum PublicEventStat
...
-        MaxMultiKill        = 14,
-        LongestImpulse      = 15,
-        LongestLife         = 16,
-        KillStreak          = 17,
-        CustomStat          = 18
+        MaxMultiKill        = 100,
+        LongestImpulse      = 101,
+        LongestLife         = 102,
+        KillStreak          = 103,
+        CustomStat00        = 200,
```
Rawaho, line 23: Are you sure this is correct?
PublicEventStats uses CustomStat as an offset for the bitmask in the packet, 200 wouldn't be correct in that scenario.

krakal: You have to do some additional processing on this. Here is how the client uses that value.

image

Rawaho: Would be good if this can be investigated further, currently setting any stat above MaxMultiKill or custom stats is broken by the PR.

Would be good to understand the exact logic the client uses when it gets the stats array and how it transforms it back into the stat indexes.
The code currently takes the statIndex from PublicEventCustomStat.tbl and adds CustomStat00 which was previously 18.
```csharp
publicEventStats.Mask.SetBit((uint)Static.PublicEvent.PublicEventStat.CustomStat00 + index, true);
```

krakal: Ah, I see the problem. There is some arithmetic that is done to translate the enum value to get the correct bit flag. Funnily, the decomplier shows machine code that takes advantage of the addition overflowing an 8-bit add to get back around rather than subtracting!
image

krakal: Think I will define the enum type-to-bit flag mapping as a dictionary rather than make a function to calculate it.

Issue #499, opened 2025-05-26
[Bug] -  Ignoring Quests of the Datachron


Val1t3:
Date: 5/26/2025
OS: Windows 11
Branch: game_rework
How To Reproduce:
- Create new character
- Execute command: !character level 50
- Click on red cross button to ignore quest call

Impossible to ignore quests in Recent Calls.

![Image](https://github.com/user-attachments/assets/a46ad5f5-9ebf-4fa9-9b5b-e3cb6c26d0ee)

Dekker3D:
I've fixed this on Omechron already. It was quite simple and I've found no issues so far. Remind me to dig it up if I forget.

Dekker3D:
Alright, submitted a PR. It seems like an easy fix, nothing unexpected.

Rawaho:
Resolved in #503

PR #503, opened 2025-06-08
Quest-ignore fix

Dekker3D:
Just copy-pasted Omechron's code for ignoring quests. It works and I see no relevant differences in the code it's calling. Fixes issue #499.


Rawaho:
QuestIgnore will move the quest to inactiveQuests if the state is Ignored or Mentioned.
Might be better to set the status first?

PR #505, opened 2025-06-23
Group - messages, Identity, renaming

krakal:
* Adds all remaining group messages
* Changes groups to use Identity/IIdentity
* Updates names to match or be neared to UI lua naming


Rawaho:
Will need to be rebased to game_rework, will look into this at some point

PR #512, opened 2025-07-11
Crowd control - messages

krakal:
* Adds messages and enum related to crowd control


Rawaho:
Small renames to allow for an easier merge of the Expedition branch which already implements CC

PR #513, opened 2025-07-13
[Bug] BuybackManager Nullptr

Ancheey:
Fix for a null pointer exception occuring when a buyback item expires but the selling player is no longer online.


Rawaho:
Thank you for the PR, nice catch!

PR #515, opened 2025-07-24
Player messages

krakal:
* Moves all player related messages/enums into .Player namespace. This is for messages sent to the player's client about themself and their interactions, made distinct from more general unit/entity related messages in the .Entity namespace that relate to any game unit
* Fixes and fills in blanks for existing player related messages
* Adds missing player related messages


Source/NexusForever.Game/Entity/ResurrectionManager.cs
```diff
@@ -99,10 +99,10 @@ public void ShowResurrection()
...
             owner.Session.EnqueueMessageEncrypted(new ServerResurrectionShow
             {
-                GhostId             = owner.ControlGuid ?? 0u,
+                GhostUnitId         = owner.ControlGuid ?? 0u,
                 RezCost             = GetCostForResurrection(),
-                TimeUntilRezMs      = 0u,
-                Dead                = true,
+                DeathPenaltyLength  = 0u,
+                PlayerIsDead        = true,
                 ShowRezFlags        = ResurrectionType,
```
Rawaho, line 106: None of these name changes seem to match the names in the client?

Source/NexusForever.Game.Abstract/Entity/IPlayer.cs
```diff
@@ -10,6 +10,7 @@
 using NexusForever.Game.Abstract.Map.Lock;
 using NexusForever.Game.Abstract.Reputation;
 using NexusForever.Game.Abstract.Social;
+using NexusForever.Game.Static.Player;
```
Rawaho, line 12: Would it make more sense to have the Player a child of the Entity namespace?
This would also allow it be expanded further for other types of entities if required.

krakal: I wouldn't put Player underneath the Entity namespace. The Player namespace is to say all these messages are useful only for the character the player is playing. With only a couple exceptions, you don't see a UnitId in these messages  because the message implicitly only applies to the player's character. Most of the state touched by these messages is actually held in the client's game state and not the unit for player's character.

Source/NexusForever.Game/Entity/ResurrectionManager.cs
krakal: GhostId to GhostUnitId = UnitId is universally how a unit is added/destroyed/referenced in the client. There is no separate GhostId table for only ghost units.

TimeUntilRezMs to DeathPenaltyLength = Named as such because the value is used to calculate fDeathPenalty in the lua table for GameLib::GetPlayerDeathState.

Dead to PlayerIsDead = Value is directly evaluated to bIsDead for the lua table for GameLib::GetPlayerDeathState. Expanded it to PlayerIsDead so that when using the message it is explicit you should only use this when referring to the player.

Rawaho: I'm more meant in regards to the last known "official" names for the properties.
Up until this point we have tried to stick as close as possible to what those names where, unless it wasn't know in which case it is guessed similar to how you are doing it with the UI.

image

Source/NexusForever.Game.Abstract/Entity/IPlayer.cs
Rawaho: Does this mean there will eventually need to be a distinction for packets and such that do effect an "unit" but are player only?
For example ServerUnitMemberOfGuildChange or ServerUnitGuildNameplateChanged?

krakal: So ServerUnitMemberOfGuildChange or ServerUnitGuildNameplateChanged are not good examples as those are for any units. For instance if someone changes their displayed guild name or joins/leaves a guild then everyone nearby needs to be sent that message for their clients to update.

To your question, I think the namespace is the primary distinction that needs to be made for packets that only affect the client's player state. Aside from that, having documentation for contributors to read to understand the difference would be a good idea which is something I was planning to do eventually.

PR #518, opened 2025-08-24
Customisation/Fix: Handle duplicate primary customisations

NeronTheTyrant:
Creating a female character with the first haircut option would always throw the same exception. This is due to there being multiple primary customisation entries for female characters with the first haircut selected.

This PR selects the first entry that matches the condition, rather than throwing if more than one entry is found.


Rawaho:
I'm happy with this fix but I think future work is required to confirm the logic for customisation as having multiple customisation enties to choose sounds incorrect.

PR #523, opened 2025-10-05
Chat messages

krakal:
* Add World.Message.Model.Chat namespace
* Add a couple messages for NPC chat
* Fills in some unknowns in existing chat message
* Adds all missing ChatFormat types


Source/NexusForever.Game/Chat/Format/Formatter/ItemGuidChatFormatter.cs
```diff
@@ -11,19 +11,19 @@ public class ItemGuidChatFormatter : IInternalChatFormatter<ChatFormatItemGuid>,
     {
         public IChatChannelTextFormatModel ToInternal(IPlayer player, ChatFormatItemGuid format)
         {
-            IItem item = player.Inventory.GetItem(format.Guid);
-            return new ChatChannelTextItemIdFormat
```
Rawaho, line 15: Having the item guid model converted to the item id model was deliberate since the chat server has no context on the item and all information needs to be in the internal message.
Now that the full item model has been added it more than likely should be that.

Source/NexusForever.Game/Chat/Format/Formatter/ItemFullChatFormatter.cs
```diff
@@ -0,0 +1,50 @@
+﻿using NexusForever.Game.Abstract.Chat.Format;
+using NexusForever.Game.Abstract.Entity;
+using NexusForever.Network.Internal.Message.Chat.Shared.Format;
+using NexusForever.Network.Internal.Message.Chat.Shared.Format.Model;
+using NexusForever.Network.World.Chat;
+using NexusForever.Network.World.Chat.Model;
+
+namespace NexusForever.Game.Chat.Format.Formatter
+{
+    public class ItemFullChatFormatter : IInternalChatFormatter<ChatFormatItemFull>, INetworkChatFormatter<ChatChannelTextItemFullFormat>
```
Rawaho, line 10: I assume there would never be a scenario where the client would send a full item model?
More than likely the converter for the item guid model should convert to either item id or item full.

Source/NexusForever.Network.Internal/Message/Chat/Shared/Format/IChatChannelTextFormatModel.cs
```diff
@@ -1,5 +1,5 @@
 ﻿using System.Text.Json.Serialization;
-using NexusForever.Game.Static.Social;
+using NexusForever.Game.Static.Chat;
 using NexusForever.Network.Internal.Message.Chat.Shared.Format.Model;
 
 namespace NexusForever.Network.Internal.Message.Chat.Shared.Format
```
Rawaho, line 5: Additional JsonDerivedType attributes required.
Currently no item link can be sent to the ChatServer because of this.

Source/NexusForever.Network.World/Message/Model/Chat/ServerChat.cs
```diff
@@ -11,16 +11,17 @@ public class ServerChat : IWritable
         public bool GM { get; set; }
         public bool Self { get; set; }
         public bool AutoResponse { get; set; }
-        public Identity From { get; set; } = new();
+        public Identity From { get; set; }
```
Rawaho, line 14: This breaks Player::SendSystemMessage, I'm assuming there are scenarios such as system messages where the from will be default.
Currently equipping or unequipping a message is broken due to this.

Rawaho (changes requested):
I was unable to test this PR due to a few major issues, see comments.
I will do a full review once those are fixed and the PR can be tested.

Source/NexusForever.Network.World/Message/Model/Chat/ServerChat.cs
krakal: My thought on Server messages that send an Identity is to not auto-construct a placeholder Identity so that users of those messages get obvious errors from the runtime rather than silently sending 0's. It encourages direct use of the already existing Identity properties attached to characters.
If the user needs to send all zeros for the Identity becuase it isn't used in that moment, I think it is good to explicitly initialise an Identity with 0's so that is clear that is what is happening rather than it happening silently.

Rawaho: I do agree, but when changing such default logic we need to ensure that existing code isn't also broken which makes the assumption about the previous defeault.

Source/NexusForever.Game/Chat/Format/Formatter/ItemFullChatFormatter.cs
krakal: Correct, the client cannot send an ItemFull formatter. In sniffs I have only seen the client send ItemId, ItemGuid, QuestId, and Loot. In theory the client can also send NavPoint and ArchiveArticle but maybe no one captured that in a sniff as those are somewhat obscure features.

For ItemId, QuestId, and Loot (and presumably NavPoint and ArchiveArticle), the server does not have to do anything special for the Client that sent the message in the ServerChatAccept message or anything special when forwarding the message on to other players.

For ItemGuid in a ClientChat message, that is always replaced when sent to other players as ItemFull. The server also needs to send ItemFull formatters in the ServerChatAccept to replace the ItemGuid placeholders in the original message on the sending client.

Source/NexusForever.Game/Chat/Format/Formatter/ItemGuidChatFormatter.cs
krakal: Short term I will revert it back to ItemId and leave ItemFull as a TODO as I don't understand enough about the new chat implementation to check what I am doing.

PR #530, opened 2025-11-06
Option messages

krakal:
* Collects existing option related messages into new Option namespace, separating messages related to the client's Option page from the Setting namespace which are game-play settings that the player cannot always change (i.e. instance difficulty)

* Adds messages for combat log options


Rawaho:
Fine with settings being renamed to options to match the client better, although I would move things such as instance difficulty to a map or instance namespace.
It probably also makes sense to update the handlers from settings to option namespace as well.

PR #534, opened 2025-11-11
Reward messages

krakal:
* Adds Rewards namespace
* Adds all RewardRotation and RewardTrack messages
* Improves the message for Premium rewards


Source/NexusForever.Game.Static/Reward/GrantedRewardFlags.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace NexusForever.Game.Static.Reward
+{
+    [Flags]
+    public enum GrantedRewardFlags : uint
+    {
+        Item     = 0x00000001,
+        Modifier = 0x00000001,
```
Rawaho: I'm assuming this value should be different?

Rawaho (changes requested):
I have made a few cleanups to the namespace, only a single issue I need confirmation on.

Source/NexusForever.Game.Static/Reward/GrantedRewardFlags.cs
krakal: It is at least as correct as what I could decipher as what I found during reverse engineering was not super clear. I don't know better names for the values but I figure that at least calling them directly something that leads to correct behaviour is just as good.

If it doesn't pass the mask test then the granted flag does not get set.

image

Rawaho: Both flags are 0x01, is that expected? If so it might be good to have a single combined value.

krakal: I have clean up the enum and renamed it to something I think makes sense.

PR #543, opened 2025-11-18
CombatLog message refinements

krakal:
* Adds missing enums for fields in CombatLog messages
* Updates CombatLog messages with enum types and names for unknown fields


Rawaho:
Just updated the locations of some of the enums to match the Expedition branch to allow for easier merging in the future.

PR #550, opened 2025-12-01
Who command musaji

Musaji:
This PR implements /who command functionality. The following commands are supported by this PR:

/who {level:int}
/who {level:int-level:int}
/who {zone:string}
/who {path:string}
/who {class:string}
/who {race:string}

/who {name:string}
/who {substring_name:string}

Where each entry has the type I placed after ":"

There are a couple of unknowns right now in regards to this implementation. I was not sure whether or not cross faction /who requests were supported in the original Wildstar. I added some code to test cross faction /who requests and a variable that toggles the ability to request these on or off. I imagine reviewers will want us to move in a single direction, so I expect this to be replaced with a hard coded support for one of the options.

I also was not sure if the requestor's name / player should show up in the who response. I followed the same protocol as I described above for factions and created a boolean toggle to turn the functionality on or off.

Additionally, during testing I was only ever able to trigger enum reads for "Combo" and "Level". That leaves 7 / 9 of the expected enum types that Krakal outlined unsupported. In order to implement support for these WhoParameter types, I will need more detailed documentation for how to trigger these enum types.

As such this PR only aims to implement 2/9 possible enum types, Level and Combo.

Tested this across multiple zones and level ranges for a collection of players on both the same faction and differing factions.

### DIFF
--- a/Source/NexusForever.Game.Static/Entity/Class.cs
+++ b/Source/NexusForever.Game.Static/Entity/Class.cs
@@ -2,6 +2,7 @@
 {
     public enum Class : byte
     {
+        None = 0,
         Warrior      = 1,
         Engineer     = 2,
         Esper        = 3,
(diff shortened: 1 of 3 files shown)


Musaji:
Closing the Pull request as I received information on how to trigger the remaining 7 who request types.

PR #552, opened 2025-12-08
Who Command Implementation

Musaji:
This PR implements the "who" commands, which were added in #529 

There are 3 WIP tags in this PR that I would like other members of the community to look at before this can be merged. 

The first tag deals with the data structure I am iterating over for the who command themselves. The PlayerManager looked like a proper thread safe alternative to my earlier IWorldSessions implementation, but it seemed as though @Rawaho was indicating I should make a separate manager for players that only updates when a relevant who query changes. 

The next tag deals with my inability to pass subzone queries via the search parameters. I have failed to do this with both the explicit query parameter "Zone" and the implicit "Combo" query parameter that attempts to parse zone data from a string. I need some guidance to get this working as I have not been able to trigger any sub zone queries so far.

The final tag deals with whether or not we should restrict string length for search queries similar to how WoW limits the string length to queries that are strictly greater than 2 characters. The client allows users to pass lower length strings so my instinct is to support all lengths of strings that the client will accept and pass to the server.

Beyond those two points, who implementation is fully supported for arbitrary groupings of AND + OR parameters. This PR supports all 9 queries that Krakal deciphered in both aggregate expressions and singular expressions.


Rawaho:
Thank you for this PR.
With recent changes to the server I thought it was better to move the who functionality, it solves the biggest issues being concurrent access.
Hopefully I have correct migrated all of the logic you implemented :D

PR #556, opened 2026-04-22
Fix physical damage mitigation to use physical offset property

onlysubs90:
This fixes the physical damage mitigation offset in DamageCalculator.

Current behavior:
- Physical damage uses DamageMitigationPctOffsetMagic
- Tech damage uses DamageMitigationPctOffsetTech
- Magic damage uses DamageMitigationPctOffsetMagic

Since DamageMitigationPctOffsetPhysical exists in Property.cs, this appears to be a copy/paste error.

This change updates physical damage mitigation to use DamageMitigationPctOffsetPhysical instead.


Rawaho:
Doesn't contain any changes.
Please review your branch and resubmit the PR.

PR #567, opened 2026-07-19
Implement chase behavior in CombatAI

xoere2:
## Summary

- `CombatAI.Update()` now moves a creature toward its current target when in combat and out of melee range, using the existing `MovementManager.Follow()` / `DirectMovementGenerator` infrastructure
- Chase paths are recalculated every 500 ms so the creature tracks a moving target without flooding the client with movement commands
- The `IsAlive`, `InCombat`, and `TargetGuid` guards ensure the path is only recalculated when the creature has a valid live target

## What this doesn't include

Spell casting is intentionally deferred. `Creature2Entry.Spell4IdActivate00–03` are interaction/activation spells (used when a player right-clicks an NPC), not combat abilities. The actual per-creature combat AI is defined via `Creature2Entry.Creature2ActionSetId → Creature2ActionEntry`, and that action set system has no game-layer implementation yet. Adding spell casting without that system would use the wrong data source.

## Test plan

- [ ] Spawn a hostile creature near a player — creature should walk toward the player once it enters combat
- [ ] Confirm creature stops when it reaches melee range (hit radii touching)
- [ ] Confirm creature re-chases if the player moves away
- [ ] Confirm creature does nothing when out of combat (threat list empty)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Source/NexusForever.Script.Main/AI/CombatAI.cs
+++ b/Source/NexusForever.Script.Main/AI/CombatAI.cs
@@ -1,4 +1,5 @@
-﻿using NexusForever.Game.Abstract.Combat;
+using System.Numerics;
+using NexusForever.Game.Abstract.Combat;
 using NexusForever.Game.Abstract.Entity;
 using NexusForever.Script.Template;
 using NexusForever.Script.Template.Filter;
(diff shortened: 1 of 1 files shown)


Rawaho:
Thank you for the PR.
Currently AI chase behaviour has already been implemented in the expedition-evil-from-the-ether branch, this branch contains a lot of PoC work for combat and will eventually be merged back into game_rework.
https://github.com/NexusForever/NexusForever/blob/146e4f0ca26f4eb4c6c5953097aeba3cef9e0829/Source/NexusForever.Script.Main/AI/CombatAI.cs#L119-L132
If you are wanting to make additional changes to the chase behaviour you should target that branch.

